PR #48, opened 2021-05-28
feat(import-data): Added deprecation notice

kidk:
Added deprecation notice to the import CLI so it's clear user should not be using this in production environments.

I also ran the linter so there were so style changes.


jbeveland27:
I wonder if we should say something like "warning: the importer is only for testing packs and not meant to be used in a production environment". Calling it deprecated feels heavy handed to me since it's still a utility we plan to use for PR validations (unless that's changed)

kidk:
Thanks @jbeveland27 I've made the changes, and wording review by @polfliet

polfliet (approved):

polfliet:
@jbeveland27 you ok with this PR?

jbeveland27:
@polfliet Ah sorry missed the update. Yep, good to go! :shipit:

jbeveland27 (approved):

PR #79, opened 2021-06-23
feat(workflow): added image validation

rudouglas:


.github/workflows/validate_images.yml
```diff
@@ -0,0 +1,22 @@
+name: Validate Images
```
aswanson-nr, line 1: I think this would fit better within the `Validate packs` workflow! That all the validations are just jobs in one workflow

utils/validate_images.js
```diff
@@ -0,0 +1,87 @@
...
+    sizeErrors = [],
+    typeErrors = [],
+    tooManyImages = [];
+    
+/** 
+* Validates an object against a JSON schema
+* @param {String} dir - The directory to parse, set by the command line argument
+* @returns {Object[]} An array of any errors found
+*/
+async function getFiles(dir) {
```
aswanson-nr, line 26: I'm not 100% sure this will work for you, but `glob` might simplify a lot of this logic. Take a look at the `validate_packs.js` script, there's a bunch of globbing going on there

.github/workflows/validate_images.yml
rudouglas: ye I was thinking that but didn't want to do it until I checked, have combined them both, let me know if that's right

utils/validate_images.js
rudouglas: Ye i was trying to use as few packages as possible but I guess it doesn't really matter if it's just for the actions, have integrated Glob now and slimmed it down but let me know if there's other places I can streamline it

utils/validate_images.js
```diff
@@ -0,0 +1,79 @@
...
+    typeErrors = [],
+    tooManyImages = [];
+    
+/** 
+* Validates an object against a JSON schema
+* @param {String} dir - The directory to parse, set by the BASE_PATH variable
+*/
+
+const globFiles = (dir) => {
+  const entries = [...glob.sync(path.resolve(dir, '**/*'))]
```
aswanson-nr, line 29: It's possible to conditionally select files with glob, so you can just pull in the images if you like! You don't have to go with this, just an idea.
```suggestion
  const entries = [...glob.sync(path.resolve(dir, '**/*.+(png|jpeg|jpg|svg)'))]
```

aswanson-nr (commented):
Since this script has a lot of output, would you mind adding one or two screenshots?

utils/validate_images.js
rudouglas: The only reason i didn't do this is because of the check to validate they are submitting images that have our allowed extensions, is there a better way to do that?

rudouglas:
Will also add screenshots

rudouglas:
The numbers here are incorrect but it will output this info to the user telling them exactly which file/folders failed the check:

Screenshot 2021-06-25 at 18 22 53
Screenshot 2021-06-25 at 18 24 28

All tests should also pass:
Screenshot 2021-06-25 at 18 31 09

rudouglas:
The validate images test currently fails because there is an image that doesn't comply:

https://github.com/newrelic/newrelic-observability-packs/blob/main/packs/nginx/icon.webp

utils/helpers.js
```diff
@@ -38,4 +40,61 @@ const readPackFile = (filePath) => path.extname(filePath) === '.json' ? readJson
...
+  return path.extname(file)
+}
+
+/** 
+* Gets an array of all files and directories in a path
+* @param {string} dir - The directory to parse, set by the BASE_PATH variable
+* @returns {Array} An array of pathnames of a globbed directory
+*/
+const globFiles = (dir) => {
+  const globbedFiles = [...glob.sync(path.resolve(dir, '**/*'))]
```
nr-kkenney, line 86: looks like this is spreading the array & putting in another. can you simplify to returning glob.sync?

nr-kkenney (commented):
there is a bit of inconsistent styling. this was actually already brought up by liz. we should add a config for consistent styling, but no need to do it here, just mentioning.

utils/__tests__/validate_images.test.js
```diff
@@ -0,0 +1,89 @@
...
+  it('Should return correct number of images', () => {
+    helpers.globFiles.mockReturnValueOnce([ 'test/path/' ]);
+    helpers.isDirectory.mockReturnValueOnce(true)
+    helpers.getImageCount.mockReturnValueOnce(10).mockReturnValueOnce(10)
+
+    validateImages()
+    expect(global.console.warn).toHaveBeenCalledTimes(2);
+  });
+
+});
```
nr-kkenney, line 89: i think you have the tests setup correctly, but the naming is a bit awkward to me. a lot of the names sound as if they are testing the helper functions, rather than calling out the image validation requirements.

as an example:
i might rename:
```
Should return correct extension
```
to
```
when given allowed image extension, does not throw / does not return error
```

utils/validate_images.js
```diff
@@ -0,0 +1,105 @@
...
+    console.warn(`\nImages should be under ${MAX_SIZE/1000000}MB:`)
+    fileSizes.map((file) => console.warn(file))
+    valid = false
+  }
+
+  if (imageCounts.length > 0) {
+    console.warn(`\nComponents should contain less than 6 images:`)
+    imageCounts.map((dir) => console.warn(dir))
+    valid = false
+  }
```
nr-kkenney, line 94: im not 100% sold on this maybe like 70/30 for/against, but i think it might be worth moving the if checks into their respective methods. that would allow you to test those methods in isolation -- each method is scoped only to its own particular check, and also prints out its own errors.

main might instead look like:
```js
  let valid = true;
  const globbedFiles = globFiles(BASE_PATH)
  const imageCountsValidationResult = validateImageCounts(globbedFiles)
  const fileSizesValidationResult = validateFileSizes(globbedFiles)
  const imageExtensionsValidationResult = validateImageExtensions(globbedFiles)

if(!result1 || !result2 || !result3){
  core.setFailed();
}
```

nr-kkenney: you might also consider moving the core.setFailed() into each method, or into its own method which idempotently sets it.

PR #80, opened 2021-06-24
Remove JS as a top-level dependency

### LINKED ISSUE #64: Automation: Remove Javascript as a top level dependency
# Summary
We do not want to require node and yarn as development dependencies, move the `package.json` and `yarn.lock` files under the `utils/` directory. We will also need to replace the Git hook that verifies that commit messages are conventional with a bash or python version. 

### Helpful links
* [Conventional commit regex](https://github.com/newrelic/newrelic-cli/blob/main/build/lint.mk#L11)
* [Commit-msg hook output example](https://github.com/newrelic/newrelic-cli/blob/main/.githooks/commit-msg)

AC:
- [ ] Move node/yarn files under the `utils/` directory
- [ ] Update the [validate packs](https://github.com/newrelic/newrelic-observability-packs/blob/main/.github/workflows/validate_packs.yml) workflow to account for the new `package.json` location
- [ ] Replace `husky` and the conventional commit hook with `shell`

zstix:
## Description
Removes the requirement for users to have `node` and `yarn` installed on their system. The only thing that was running on contributor's machines was _Husky_ (for enforcing conventional commits). This PR moves the yarn-related files into the `utils/` directory.

## To Do
* [x] Move JS files
* [x] ~Add in Hook~
* [x] Add check in PR validation workflow

## Related Issue(s)
* Closes #64


zstix:
The remaining work that needs to be done is to run our script `utils/conventional-commits.sh` to check if all the commits for a PR are following the correct convention.

To get a list of the recent commits (with, _just_ the commit message), you can run:

```
git log origin..HEAD --pretty=format:%s
```

zstix:
It turns out that we can't check in files in the `.git/hooks` directory. This means that we are unable to set up the `commit-msg` hook and check it into the repo.

We've decided that we will _not_ check for commit message validity when the users is working on their machine and will _only_ be checking during PR validation.

aswanson-nr:
I moved the script from `sh` to `bash` because arrays are far easier to deal with in bash and we no longer need this to run on any random machine.

rudouglas (approved):
LGTM

PR #82, opened 2021-06-24
Add labeling workflow

LizBaker:
## Summary

Adds a workflow to auto label PRs (that directly edit a pack) with their support level. Also adds a label for new packs

Closes https://github.com/newrelic/newrelic-observability-packs/issues/66

Screenshots from a successful test:
Screen Shot 2021-06-25 at 4 27 54 PM
Screen Shot 2021-06-25 at 4 28 17 PM


rudouglas (dismissed):
LGTM

.github/workflows/labels.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Install dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile
+
+      - name: Check support level
+        id: support
+        run: |
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
+          yarn check-support-level $URL
```
nr-kkenney, line 36: i think the yarn steps are gonna need cd utils, since we moved the javascript stuff into the utils folder.

utils/__tests__/check_support_level.test.js
```diff
@@ -0,0 +1,98 @@
...
+  fs.readFileSync.mockReturnValueOnce(yml);
+};
+
+const STATUS = {
+  ADDED: 'added',
+  MODIFIED: 'modified',
+  REMOVED: 'removed',
+};
+
+describe('Action: Check support level', () => {
```
nr-kkenney, line 43: i remember us mentioning multiple support level labels being added. maybe there should be a test for that as well.

LizBaker: so true! good idea

nr-kkenney (changes requested):
i dropped the comments on the specific spots 🙈

.github/workflows/labels.yml
nr-kkenney: i also dropped this comment on ruairi's pr. if you feel like using this:
i think there is a default you can set for working directory.

something like:
```yml
jobs:
  x-job:
    defaults:
      working-directory: utils
```

so that becomes the default directory when commands execute, and you dont need cd utils for every step.
im not 100% thats correct, but it should be something like that.
also dont feel like you have to, just dropping this an option.

nr-kkenney (dismissed):

nr-kkenney (approved):

PR #84, opened 2021-06-28
feat(schema): Updating main configuration schema

rudouglas:


rudouglas:
This fails because one of the packs configs does not have a `description` but it is in an `/instrumentation/logging/` folder which I don't think we are accounting for yet in our validation

https://github.com/newrelic/newrelic-observability-packs/blob/36b292e7248d7226065458351de4812ba994bac4/packs/elasticsearch/instrumentation/logging/elasticsearch.yml

PR #85, opened 2021-06-28
chore: replace nginx icon with png

rudouglas:


nr-kkenney (changes requested):
you will also need to change `config.yml`
this line in particular:
```none
icon: icon.webp
```

rudouglas:
Duplicate PR

Issue #87, opened 2021-06-28
add prettier config for formatting consistency


nr-kkenney:
### Acceptance Criteria:
- [ ] add .prettierrc.js & ignore file

PR #88, opened 2021-06-28
add prettier config & ignore, format files

nr-kkenney:
PR to:
* add prettier config & ignore file.
* format files in the utils directory.

Resolves: #87


rudouglas (approved):
LGTM

PR #96, opened 2021-06-30
feat(workflow): image validation and tests

rudouglas:
Screenshot 2021-06-30 at 13 50 55

### DIFF
--- a/.github/workflows/validate_packs.yml
+++ b/.github/workflows/validate_packs.yml
@@ -49,3 +49,20 @@ jobs:
       - name: Check for non-conventional commits
         run: ./utils/conventional-commits.sh
 
+  ensure-images-are-valid:
+    name: Ensure images are valid
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 12
+
+      - name: Install dependencies
+        run: cd utils && yarn install
+
+      - name: Validate Images
+        run: cd utils && yarn validate-images
--- a/utils/__tests__/validate_images.test.js
+++ b/utils/__tests__/validate_images.test.js
@@ -1,13 +1,15 @@
 'use strict';
-const { checkImageCounts, checkImageExtensions, checkFileSizes } = require('../validate_images');
+const {
+  validateImageCounts,
+  validateImageExtensions,
+  validateFileSizes,
+} = require('../validate_images');
 const helpers = require('../helpers');
-const glob = require('glob');
-const path = require('path');
-const isImage = require('is-image');
-const fs = require("fs");
+const core = require('@actions/core');
+const fs = require('fs');
 
-
-jest.mock('fs')
+jest.mock('@actions/core');
+jest.mock('fs');
 jest.mock('glob');
 jest.spyOn(global.console, 'warn').mockImplementation(() => {});
 jest.mock('../helpers', () => ({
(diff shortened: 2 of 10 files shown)


.github/workflows/validate_packs.yml
```diff
@@ -49,3 +49,20 @@ jobs:
       - name: Check for non-conventional commits
         run: ./utils/conventional-commits.sh
 
+  ensure-images-are-valid:
```
nr-kkenney, line 52: i think there is a default you can set for working directory.

something like:
```yml
jobs:
  x-job:
    defaults:
      working-directory: utils
```

im not 100% thats correct, but it should be something like that.

nr-kkenney: just a heads up, no need to change this. just giving some insight.

utils/validate_images.js
```diff
@@ -0,0 +1,109 @@
...
+
+const main = () => {
+  const globbedFiles = globFiles(BASE_PATH);
+  const imageCounts = checkImageCounts(globbedFiles);
+  const fileSizes = checkFileSizes(globbedFiles);
+  const imageExtensions = checkImageExtensions(globbedFiles);
+
+  if (!imageCounts || !fileSizes || !imageExtensions) {
+    process.exit(1);
+  }
```
nr-kkenney, line 97: you dont need this step. if core.setFailed is called, that will set the exit code to 1.
https://github.com/actions/toolkit/blob/2f164000dcd42fb08287824a3bc3030dbed33687/packages/core/src/core.ts#L108

utils/validate_images.js
```diff
@@ -0,0 +1,109 @@
...
+    .filter((file) => {
+      return !ALLOWED_IMG_EXT.includes(getFileExtension(file));
+    });
+  if (extensions.length > 0) {
+    core.setFailed(`Images should be of format ${[...ALLOWED_IMG_EXT]}:`);
+    console.warn(`\nPlease check the following images:`);
+    extensions.map((file) => console.warn(file));
+    return false;
+  }
+  return true;
```
nr-kkenney, line 86: dont need to return anything if we no longer need to set process code at the end. see below comment.

utils/validate_images.js
```diff
@@ -0,0 +1,109 @@
...
+};
+
+const main = () => {
+  const globbedFiles = globFiles(BASE_PATH);
+  const imageCounts = checkImageCounts(globbedFiles);
+  const fileSizes = checkFileSizes(globbedFiles);
+  const imageExtensions = checkImageExtensions(globbedFiles);
```
nr-kkenney, line 93: might consider renaming the functions. 
so:
checkImageCounts -> validateImageCounts
checkFileSizes -> validateFileSizes
checkImageExtensions -> validateImageExtensions

utils/validate_images.js
```diff
@@ -0,0 +1,109 @@
...
+        size: `${getFileSize(file) / 1000000}MB`,
+      };
+    });
+  if (sizes.length > 0) {
+    core.setFailed(`Images should be under ${MAX_SIZE / 1000000}MB:`);
+    console.warn(`\nPlease check the following images:`);
+    sizes.map((file) => console.warn(file));
+    return false;
+  }
+  return true;
```
nr-kkenney, line 66: same comment as below.

utils/validate_images.js
```diff
@@ -0,0 +1,109 @@
...
+        imageCount: getImageCount(folder),
+      };
+    });
+  if (directories.length > 0) {
+    core.setFailed('Components should contain less than 6 images');
+    console.warn(`\nPlease check the following directories:`);
+    directories.map((dir) => console.warn(dir));
+    return false;
+  }
+  return true;
```
nr-kkenney, line 40: same comment as below.

utils/__tests__/validate_images.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+  test('Given a file path, returns the extension of the file', () => {
+    const input = 'test/file/ext.png'
+    expect(helpers.getFileExtension(input)).toBe('.png');
+  });
+
+  test('Given a directory path, returns empty for file extension', () => {
+    const input = 'test/file/ext'
+    expect(helpers.getFileExtension(input)).toBe('');
+  });
```
nr-kkenney, line 36: i think these tests should go into a helpers.test.js

utils/__tests__/validate_images.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+  test('Given allowed image size, does not throw an error', () => {
+    isImage('test/path/config.png')
```
nr-kkenney, line 39: i dont think you need this isImage line

utils/__tests__/validate_images.test.js
```diff
@@ -0,0 +1,88 @@
...
+    helpers.getFileSize.mockReturnValueOnce(1000).mockReturnValueOnce(1000)
+    helpers.isDirectory.mockReturnValueOnce(true)
+    fs.statSync.mockReturnValueOnce('test/path/config.yml')
+
+    checkFileSizes(globMockSize)
+    expect(global.console.warn).not.toHaveBeenCalled();
+  });
+
+  test('Given allowed image size, throws an error', () => {
+    isImage('test/path/config.png')
```
nr-kkenney, line 49: i dont think you need this isImage line

utils/__tests__/validate_images.test.js
```diff
@@ -0,0 +1,88 @@
...
+  test('Given allowed number of images in a directory, does not throw an error', () => {
+    const globMock = [ 'test/path/config' ];
+    helpers.isDirectory.mockReturnValueOnce(true)
+    helpers.getImageCount.mockReturnValueOnce(2).mockReturnValueOnce(2)
+
+    checkImageCounts(globMock)
+    expect(global.console.warn).not.toHaveBeenCalled();
+  });
+
+  test('Given allowed number of images in a directory, throws an error', () => {
```
nr-kkenney, line 79: this is a comment for all the test names:
if you could change up the first part of the name "given . . ." so that they are different for the positive case & the negative case. i also have a personal preference to include the unit under test in the name in some way. that could be in the test, or the describe block.

so for these two tests, i might do:
validateImageCount, given <= 6 images in a directory, does not error (passes)
validateImageCount, given >6 images in a directory, throws error (fails)

utils/validate_images.js
```diff
@@ -0,0 +1,100 @@
...
+  isDirectory,
+} = require('./helpers');
+
+const BASE_PATH = '../packs/';
+const MAX_SIZE = 4000000;
+const MAX_NUM_IMG = 6;
+const ALLOWED_IMG_EXT = ['.png', '.jpeg', '.jpg', '.svg'];
+
+/**
+ * Validate all folders contain no more than MAX_NUM_IMG images
```
nr-kkenney, line 18: this is something you dont need to address, and just my opinion. you might describe the side effect that happens when you call the function & validation fails in the description / summary section of the doc.

so for this, and the other functions, i might write - 'Validate all folders contain no more than MAX_NUM_IMG images. if validation fails, process.exitCode is set to 1.'

nr-kkenney (approved):

PR #74, opened 2021-06-15
chore(spec): add spec docs for config, alerts, and dashboards

jbeveland27:
Also fixed dashboard validation issues in all packs

PR #97, opened 2021-06-30
chore: update schema for main config

rudouglas:
Also updated some tests that where dependent on this


jbeveland27:
Hey @rudouglas 👋🏼 #74 has some similar work to what you've done here. Might be worth syncing up to see if we should combine one into the other. I think @jpvajda is going to discuss some of that at today's standup.

jpvajda:
Our decision after syncing was to: close this PR as @jbeveland27 accounting for this work in #74, he was going to add some of the code used in this PR to his.

PR #98, opened 2021-06-30
chore: contributing & template updates

jpvajda:
This PR updates the contributing guide as well as makes a few modifications to the pack template.


jpvajda:
@jbeveland27 I would love a review on this from you if you have time.

CONTRIBUTING.md
```diff
@@ -11,26 +28,118 @@
...
+your interactions with the project and its participants.
 
 ## Observability Pack Guidelines
 
-We encourage all contributors to actively engage in the creation and maintenance of Observability Packs. Whether you work at New Relic or use New Relic as a customer, the community is open to your expertise!
+We encourage all contributors to actively engage in the creation and maintenance of
+Observability Packs. Whether you work at New Relic or use New Relic as a customer, the community is open to your expertise!
+
+- `Step 1`: Review the [pack template config](./_template/config.yml) file for a definition of how to create a pack.
+- `Step 2`: Review the [Pack Config File Schema](https://github.com/newrelic/newrelic-observability-packs/blob/f3acfd3efc34899c088c01b48c2e2f2210dd2657/docs/config-spec.md) for structure and limits you need to consider.
```
jbeveland27, line 49: This link should be to the main branch rather than a commit (which will end up out of date)
```suggestion
- `Step 2`: Review the [Pack Config File Schema](https://github.com/newrelic/newrelic-observability-packs/blob/main/docs/config-spec.md) for structure and limits you need to consider.
```

CONTRIBUTING.md
```diff
@@ -11,26 +28,118 @@
...
+### Website URLs
 
-If you decide to create a new Observability Pack or review a PR please keep the following in mind.
+> a reference to a website relating tot he pack.
 
-* Your new pack must have a unique name
-  * The `Validate Packs / Ensure pack names are unique` job handles verifying that there are no collisions when you create a pull request
-  * **Note:** Names are stripped of any punctuation and white space is replaced by `-` before doing any comparisons
+- Optional
+- Valid URL only https://www.newrelic.com
```
jbeveland27, line 142: My $0.02 - this should be a link to the `main_config.md` doc rather than being duplicated info that could go stale as the doc gets updates. I think the `docs` directory should be our source of truth as much as possible. 

It might be worth copying some of the info that's here into `main_config.md`.

_template/config.yml
```diff
@@ -1,23 +1,22 @@
 
-# Name of the pack
+# Name of the pack (required)
 name: Template pack
```
jbeveland27, line 3: For compatibility with the APIs and usage in the CLI, it'd be worthwhile to establish hyphenated lowercase as the format for names
```suggestion
name: template-pack
```

CONTRIBUTING.md
```diff
@@ -11,26 +28,118 @@
...
+- Required
+- No limit to the amount of authors
+- The `TBD` job verifies pack level
+
+### Optional field best practices
+
+Other fields in a pack are optional but it's recommended you consider the following to offer the best possible pack
+experience for users.
+
+#### Short Description
```
jbeveland27, line 101: This is `summary` now (see [config-spec.md](https://github.com/newrelic/newrelic-observability-packs/blob/main/docs/config-spec.md))

CONTRIBUTING.md
jpvajda: that's a good point.. I wasn't sure how much you wanted to make the `config-spec.md` a how to guide. So the intention here was to step a user through what to think about when creating a pack. I'm happy to transfer a lot of this content to that file if that is where you'd prefer to see it.

jbeveland27: I think much of this is already encapsulated in the spec file (required/optional, limits, format, etc.). Main concern is if it's represented here, it's just one more thing to be mindful of if the spec is modified.

jbeveland27: But if you find value in keeping it here for contributors, that's cool too

jpvajda: 👍

jpvajda:
@jbeveland27 alrighty, I resolved all the feedback and opted to keep a more detailed set of instructions in this file for now to aid contributors in getting started.

jbeveland27 (approved):
🚢 🇮🇹

PR #107, opened 2021-07-15
Update package.json

LizBaker:
## Summary

We moved the package.json into the utils folder but one yarn command still had a path with /utils which caused the action to break


aswanson-nr (approved):

jpvajda:
👏

jpvajda:
the support labels issue is because we moved the package.json at one point but the file path to the script didnt get updated, @LizBaker fixed it. 🥳

PR #143, opened 2021-08-18
Config spec update

jpvajda:
This PR updates the configs for alerts, dashboard, and packs to highlight new functionality added to nerdgraph. It also updates the contributing guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,6 @@
       - [Summary](#summary)
     - [Optional field best practices](#optional-field-best-practices)
       - [Title](#title)
-      - [PackUrl](#packurl)
     - [Images or screenshots](#images-or-screenshots)
       - [Image recommended for best results](#image-recommended-for-best-results)
     - [Logos](#logos)
@@ -114,12 +113,6 @@ experience for users.
 #### Title
 
 > an option field used when the pack is referenced.
-
-- `Optional`
-
-#### PackUrl
-
-> URL to link to the location of this Observability Pack in the open source repo
 
 - `Optional`
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,13 +9,17 @@
       - [Description](#description)
       - [Level](#level)
       - [Author](#author)
-    - [Optional field best practices](#optional-field-best-practices)
       - [Summary](#summary)
+    - [Optional field best practices](#optional-field-best-practices)
+      - [Title](#title)
+      - [PackUrl](#packurl)
     - [Images or screenshots](#images-or-screenshots)
       - [Image recommended for best results](#image-recommended-for-best-results)
     - [Logos](#logos)
       - [logo recommendations or best results](#logo-recommendations-or-best-results)
+    - [Icons](#icons)
     - [Website URLs](#website-urls)
+    - [Image requirements](#image-requirements)
   - [Feature Requests](#feature-requests)
   - [Pull Requests](#pull-requests)
     - [Status Checks](#status-checks)
@@ -67,6 +71,7 @@ When creating a new Observability Pack or reviewing a PR please keep the followi
 - Avoid PR collisions by checking if any [open PRs](https://github.com/newrelic/newrelic-observability-packs/pulls) are using the same name you wish to use.
 - Pack names are stripped of any punctuation and white space is replaced by `-` before doing any comparisons.
 
+
 #### Description
 
 > A detailed description of the pack and why it's useful.
@@ -87,24 +92,36 @@ When creating a new Observability Pack or reviewing a PR please keep the followi
 
 #### Author
 
-> The creator(s) of the pack
+> The creator(s) or contributors of the pack
 
 - `Required`
 - No limit to the amount of authors
 - A Github Action job verifies pack level
+
+#### Summary
+
+> A brief summary of the pack functionality.
+
+- `Required`
+- Plain text only
+- limited to 250 characters
 
 ### Optional field best practices
 
 Other fields in a pack are optional but it's recommended you consider the following to offer the best possible pack
 experience for users.
 
-#### Summary
+#### Title
 
-> A brief summary of the pack functionality.
+> an option field used when the pack is referenced.
 
 - `Optional`
-- Plain text only
-- limited to 250 characters
+
+#### PackUrl
+
+> URL to link to the location of this Observability Pack in the open source repo
+
+- `Optional`
 
 ### Images or screenshots
 
@@ -133,6 +150,13 @@ experience for users.
 
 - Aspect ratio: 1:1
 - 250px (width) x 100px (height)
+
+### Icons
+
+- `Optional`
+- Not currently used
+- .png or .jpeg or .svg format
+- Max 1
 
 ### Website URLs
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,7 +77,7 @@ When creating a new Observability Pack or reviewing a PR please keep the followi
 > A detailed description of the pack and why it's useful.
 
 - `Required`
-- Plain text only
+- Markdown or plain text
 - limited to 2000 characters
 - A Github Action verifies pack descriptions
 
--- a/docs/alert-spec.md
+++ b/docs/alert-spec.md
@@ -16,19 +16,23 @@ packs/<pack_name>/alerts/static-alert.yml
 
 ```yaml
 ####################################################################################
-# Metadata - these fields will be removed from the AlertsNrqlConditionInput 
-# definition
+# Metadata - these fields will be removed from the AlertsNrqlConditionInput
 ####################################################################################
 
-# A short form description for this Alert condition
+# the name of the Alert
+name: string, required
+
+# A short description for this Alert condition
 details: string, optional
 
-# Type of alert condition
+# Type of Alert condition
 type: enum, required # One of [ BASELINE, OUTLIER, STATIC ]
 
+# Url of the Alert in this pack
+url: string, required
 
 ####################################################################################
-# Alert Condition Definition - not validating this, as the type is not controlled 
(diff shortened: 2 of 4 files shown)


mbazhlekova (approved):
Looks good to me! 🚀

docs/dashboard-spec.md
```diff
@@ -14,6 +14,31 @@ packs/<pack_name>/dashboards/apache.json
...
+# Metadata
+####################################################################################
+
+# the name of the Dashboard
+name: string, required
+
+# A short description for this Dashboard
+description: string, required
+
+# List of screenshot URLs related to this Dashboard
```
aswanson-nr, line 29: This comment needs to be updated

docs/dashboard-spec.md
```diff
@@ -14,6 +14,31 @@ packs/<pack_name>/dashboards/apache.json
...
+type: list, required
```
aswanson-nr, line 30: This field is currently not required in the json schema

docs/config-spec.md
```diff
@@ -17,73 +17,68 @@ For example:
 # Metadata
 #####################
 
-# TBD
-# Unique identifier for this observability pack
-# uuid: guid, required
+# A unique Observability Pack identifier
+uuid: guid, required
```
aswanson-nr, line 21: This is not currently part of the json schema.

docs/config-spec.md
```diff
@@ -17,73 +17,68 @@ For example:
...
+
+# The name of the pack displayed everywhere the pack is referenced
+title: string, optional
```
aswanson-nr, line 24: Should this be required? We will be using it to display the pack in the both catalogs.

docs/config-spec.md
```diff
@@ -17,73 +17,68 @@ For example:
...
 description: string, required
 
-# TBD
-# url slug for linking on public website
-# recommended format: make this a lowercase and hyphenated version
-# of the `name` field
-# slug: string, required
+# Displayed in search results and recommendations. Summarizes a packs functionality.
+# A short form description for this Observability Pack.
+summary: string, required
```
aswanson-nr, line 36: This is not currently required in the json schema

docs/config-spec.md
```diff
@@ -17,73 +17,68 @@ For example:
...
+
+# URL to link to the location of this Observability Pack in the open source repo
+packUrl: string, optional
```
aswanson-nr, line 39: This is purely a field in the GraphQL API, it will not be present in the pack configuration.

docs/alert-spec.md
```diff
@@ -16,19 +16,23 @@ packs/<pack_name>/alerts/static-alert.yml
...
+
+# A short description for this Alert condition
 details: string, optional
 
-# Type of alert condition
+# Type of Alert condition
 type: enum, required # One of [ BASELINE, OUTLIER, STATIC ]
 
+# Url of the Alert in this pack
+url: string, required
```
aswanson-nr, line 32: What would be the value of this url? I'm not sure what it would point to. Also we may want to avoid arbitrary URLs in these configurations because they can leave you open to XSS attacks.

CONTRIBUTING.md
```diff
@@ -87,24 +92,36 @@ When creating a new Observability Pack or reviewing a PR please keep the followi
...
 
 - `Required`
 - No limit to the amount of authors
 - A Github Action job verifies pack level
 
+#### Summary
+
+> A brief summary of the pack functionality.
+
+- `Required`
```
aswanson-nr, line 105: This is not currently required in the json schema

docs/dashboard-spec.md
```diff
@@ -14,6 +14,31 @@ packs/<pack_name>/dashboards/apache.json
...
+
+# URL of Dashboard JSON definition
+url: string, required
```
aswanson-nr, line 33: Same comment as the Alerts URL.

aswanson-nr (commented):
Sorry for spamming a bunch of comments! I'm totally fine with merging this without updating the json schema as long as we have work defined to go in and update the schema and add the required fields to all packs.

docs/dashboard-spec.md
jpvajda: how so? What's the update?

docs/dashboard-spec.md
jpvajda: It shows are requires in nerdgraph though...


Screen Shot 2021-08-19 at 8 29 08 AM

jpvajda: hol up, this means the field is

`String!` means that the field is non-nullable, meaning that the GraphQL service promises to always give you a value when you query this field. In the type language, we'll represent those with an exclamation mark.

I'll fix it. I misunderstood the `!`

jpvajda: I'm confirming what this actually means as it relates to this field.

docs/config-spec.md
jpvajda: What is this then? 


Screen Shot 2021-08-19 at 8 47 44 AM

docs/config-spec.md
jpvajda: I would say yes to being required for the title field, you make a good point.  it's currently set to nullable in graphql 

Screen Shot 2021-08-19 at 8 49 16 AM

docs/config-spec.md
jpvajda: This is currently listed as a non-nullable field in graphql

Screen Shot 2021-08-19 at 8 50 46 AM

docs/config-spec.md
jpvajda: gotcha.. so a user won't be contributing this value. I can remove it.

jpvajda: actually I'd like to keep this documented for now, until we get our guides in place and the json schema updated. but i'll add a comment clarifying

docs/alert-spec.md
jpvajda: Not sure, it's in Graphql though the comment says URL of Alert JSON definition

Screen Shot 2021-08-19 at 9 00 23 AM

docs/dashboard-spec.md
jpvajda: @rudouglas did confirm this `string!` does mean required..

CONTRIBUTING.md
jpvajda: it is in graphql though...

Screen Shot 2021-08-19 at 9 11 05 AM

docs/dashboard-spec.md
jpvajda: it's defined as URL of Dashboard JSON definition

Screen Shot 2021-08-19 at 9 12 03 AM

docs/dashboard-spec.md
aswanson-nr: That comment refers to a list of URLs, but it's describing the `description` field

jpvajda:
all the feedback was resolved, so I feel it's :ok:  to merge this in for reference as we finish this project up, then we can resolve this source of truth discrepancy with this issue before GA, removing these config files .

docs/config-spec.md
jbeveland27: IMO this doesn't belong here. A user will not enter this and it's misleading to say it's a part of the config spec. It also shouldn't be a part of the contributing guide - user's shouldn't be submitting this as valid pack input, it will just be ignored.

CONTRIBUTING.md
```diff
@@ -9,13 +9,17 @@
       - [Description](#description)
       - [Level](#level)
       - [Author](#author)
-    - [Optional field best practices](#optional-field-best-practices)
       - [Summary](#summary)
+    - [Optional field best practices](#optional-field-best-practices)
+      - [Title](#title)
+      - [PackUrl](#packurl)
```
jbeveland27, line 15: Don't think this belongs in the contributing guide (see other comment as well in the config-spec.md file)

docs/alert-spec.md
jbeveland27: `url` doesn't belong in the source alert spec - it's something added by the service and will be ignored if the user supplies it.

docs/dashboard-spec.md
```diff
@@ -14,6 +14,30 @@ packs/<pack_name>/dashboards/apache.json
...
+####################################################################################
+
+# the name of the Dashboard
+name: string, required
+
+# A short description for this Dashboard
+description: string, required
+
+# List of screenshot URLs related to this Dashboard
+type: list, required
```
jbeveland27, line 29: `type` is not the field, `screenshots` is. Also, this field isn't actually needed as the service just adds it.

docs/dashboard-spec.md
jbeveland27: Same thing here - the service adds this URL field automatically. We shouldn't be asking users to enter it.

jbeveland27 (changes requested):

docs/config-spec.md
jpvajda: I removed it.

docs/alert-spec.md
jpvajda: I'll remove it.

docs/dashboard-spec.md
jpvajda: I'll remove it.

CONTRIBUTING.md
jpvajda: Removed.

docs/dashboard-spec.md
jpvajda: Will remove.

docs/config-spec.md
aswanson-nr: @jpvajda Looks like this got re-added!

jpvajda: 💇  it's gone now.

jbeveland27 (approved):
🚀 🚢

PR #146, opened 2021-08-18
Add doc and schema for documentation field

roadlittledawn:
## Summary

Updates doc and JSON schema for forthcoming pack `documentation` field.

## Notes
* Held off on URL validation for now as there isn't a great, built-in way to do that in JSON schema. [Regex has it's pitfalls](https://json-schema.org/understanding-json-schema/reference/string.html#index-12) and [this](https://www.stephenlewis.me/notes/json-schema-url-validation/) is closest way I found, which felt like it could hit some snags.

## Related issues / PRs
https://github.com/newrelic/newrelic-observability-packs/issues/140
open-install-library-service/pull/56

### DIFF
--- a/utils/schemas/main_config.json
+++ b/utils/schemas/main_config.json
@@ -41,7 +41,12 @@
       ],
       "icon": "icon.jpg",
       "logo": "logo.png",
-      "website": "https://www.newrelic.com"
+      "website": "https://www.newrelic.com",
+      "documentation": {
+        "name": "How to use this pack",
+        "url": "https://example.com",
+        "description": "A brief summary of what this doc entails."
+      }
     }
   ],
   "required": [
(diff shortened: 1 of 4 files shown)


utils/schemas/main_config.json
```diff
@@ -370,6 +375,77 @@
       "examples": [
         "https://www.newrelic.com"
       ]
+    },
+    "documentation": {
+      "$id": "#/properties/documentation",
+      "type:": "array",
```
nr-kkenney, line 381: extra colon in the field name.

utils/schemas/main_config.json
```diff
@@ -370,6 +375,77 @@
...
+                "$id": "#/properties/documentation/items/anyOf/0/properties/url",
+                "type": "string",
+                "title": "The url schema",
+                "description": "The URL for the documentation resource.",
+                "default": "",
+                "examples": [
+                  "https://example.com"
+                ]
+              },
+              "description": {
```
nr-kkenney, line 435: i think john mentioned limiting this field to ~250 characters.

nr-kkenney (changes requested):
left a couple comments. another general thing is that we should update the tests with a few new cases to validate the new fields.

utils/schemas/main_config.json
aswanson-nr: I think that was summary field, description is going to stay long form

utils/schemas/main_config.json
roadlittledawn: 🤦  thanks! i'm now going to make an eye appointment ;)

nr-kkenney (approved):
lgtm

utils/schemas/main_config.json
```diff
@@ -347,29 +299,86 @@
...
+          }
+        ]
+      ],
+      "items": {
+        "$id": "#/properties/documentation/items",
+        "anyOf": [
+          {
+            "$id": "#/properties/documentation/items/anyOf/0",
+            "type": "object",
+            "title": "The first anyOf schema",
```
aswanson-nr, line 342: I think this might be left over from copy paste!

aswanson-nr (changes requested):
Saw one comment that looks like it needs to be updated, the rest LGTM!

utils/schemas/main_config.json
roadlittledawn: which part? the `id` is unique. and i just took the generic `title` value like the others

aswanson-nr: Gotcha! could you just update the `description`? Maybe something like "a collection of name, url, and optional description for a piece of documentation"

roadlittledawn: added that just now

PR #151, opened 2021-08-19
This is a test PR

jpvajda:
seeing if this works. :D I though we changed description field to support markdown, but our [contributing documentation](https://github.com/newrelic/newrelic-observability-packs/blob/main/CONTRIBUTING.md#description) is saying this field is plain text only.


jpvajda:
we confirmed markdown is supported on  only on the pack details page, not in the list view.

PR #154, opened 2021-08-20
Update image validation check

LizBaker:
## Summary

Updates the image count validation function to count images per nested pack and ignore icon/logo files based on their paths found in the config

Updates tests


utils/validate_images.js
```diff
@@ -18,19 +20,37 @@ const ALLOWED_IMG_EXT = ['.png', '.jpeg', '.jpg', '.svg'];
...
+      }
     })
-    .map((folder) => {
-      return {
-        folder,
-        imageCount: getImageCount(folder),
-      };
-    });
-  if (directories.length > 0) {
+    .filter(Boolean);
```
nr-kkenney, line 51: what does this filter do?

LizBaker: the map will loop over the packs and make an entry for each one even if that entry is undefined (in the case that the image count is not exceeding 6) so this filter quickly drops any item in the array that's falsy, all the undefineds

nr-kkenney (approved):
asked a question since i wasn't sure about a line, but looks good.

PR #155, opened 2021-08-23
Update support label script

LizBaker:
## Summary
Updates the support label script and it's tests for new nested file structure

### Related Issues
* Closes https://github.com/newrelic/newrelic-observability-packs/issues/138


utils/check_support_level.js
```diff
@@ -34,29 +38,32 @@ const getNextLink = (linkHeader) => {
...
     }
-    return acc;
-  }, []);
+  });
+
+  const configFilesToRead = configFiles
+    .map((config) => {
+      const configRelativePath = removeRepoPathPrefix(config);
+      const regex = new RegExp(`^${path.dirname(configRelativePath)}.*`);
+      const fileMatch = files.find((file) => file.filename.match(regex));
```
zstix, line 56: If we're trying to determine if a file is the same, could we just use [`path.basename`](https://nodejs.org/api/path.html#path_path_basename_path_ext)?

```suggestion
      const fileMatch = files.find((file) => file === path.basename(config));
```

aswanson-nr: This is checking to see if a file exists within a quickstart. 
So if I had changes like:
```
packs/apache/dashboards/screenshot1.png - Added
```
and I have a list of pack directories:
```
packs/apache
packs/browser
...
```
This allows us to confirm that the new screen shot is part of the `apache` pack and to then read the `apache/config.yml` to check the support level of that pack.

zstix: Ahh gotcha. Sorry that wasn't obvious to me as I was scanning through the code. Thanks!

zstix (approved):

PR #158, opened 2021-08-24
add validation for ensuring icon and logo exist (when supplied)

nr-kkenney:
## Summary

* added a script for validating icons and logos exist if supplied
* added tests for script
* script is invoked as a job on the validate_packs workflow

### DIFF
--- a/utils/validate_icons_and_logos.js
+++ b/utils/validate_icons_and_logos.js
@@ -1,0 +1,69 @@
+'use strict';
+const fs = require('fs');
+
+const { readPackFile, findMainPackConfigFiles } = require('./helpers');
+
+/**
+ * Method to validate icons and logos exist if supplied.
+ * @returns {string[]} error messages for errors encountered
+ */
+const validateIconAndLogo = () => {
+  const errorMessages = [];
+
+  for (const configPath of findMainPackConfigFiles()) {
+    var config = readPackFile(configPath);
+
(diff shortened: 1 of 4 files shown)


utils/validate_icons_and_logos.js
```diff
@@ -0,0 +1,75 @@
...
+const { readPackFile, findMainPackConfigFiles } = require('./helpers');
+
+/**
+ * Method to validate icons and logos exist if supplied.
+ * @returns {string[]} error messages for errors encountered
+ */
+const validateIconAndLogo = () => {
+  const errorMessages = [];
+
+  for (const configPath of findMainPackConfigFiles()) {
```
aswanson-nr, line 14: If you pass in the return value of `findMainPackConfigFiles`, it would probably make it easier to test. You won't need to stub it out since you can just pass in an array of file paths.

aswanson-nr (approved):
Looks fantastic! I had one suggestion that might make testing easier, but it's totally optional

utils/validate_icons_and_logos.js
nr-kkenney: do you mean like:
```js
const configPaths = findMainPackConfigFiles();

for(const configPath of configPaths){
}
```

i think i still need to mock the output of the method. this was also the way i originally had it and just had a thought to combine the lines 🙈

PR #183, opened 2021-09-01
feat(install): added installation plans

kidk:
## WIP: Added installation plans

Working on adding all the current add more data sources to the repository based on the spec defined by @jbeveland27

### DIFF
--- a/install/guided-install.yml
+++ b/install/guided-install.yml
@@ -1,0 +1,14 @@
+---
+id: guided-install
+name: Guided install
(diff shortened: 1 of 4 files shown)


jbeveland27:
Note: I removed the `name` field from the schema, as it was redundant having both `name` and `title` (and it wasn't clear that we'd use both in the UI).

jbeveland27:
In #203, I added some initial install plans as well and nested them under well-named directories. I think we could follow that pattern here.

jbeveland27:
Closing in favor of #220

PR #193, opened 2021-09-03
feat(Emil Hammarstrand): Updated Keywords In Config Files

EmilHammarstrand:
## Summary

Updated all the keywords in the config files. 
Packs that could not be updated due to not finding them in the Excel file was:

_Apis/* Redhat/*
Oracle/oracle-database
Lamp
Postgreql
Prometheus-node-export-macos_


polfliet:
Hi @EmilHammarstrand, we are getting closer but we really only want to see changes to the keywords field.
Ideally we see a nice diff in the PR that only shows adding the keywords field.
Right now, it indicates changes to the description, author and other fields as well. Maybe your script is reading and overwriting those fields again? If you can please change that, so it only adds the keywords field that would make it nice and clean to review. Thanks!

polfliet:
Closing this PR as there is some more work today before the team can review it

PR #194, opened 2021-09-03
Updated Keywords

EmilHammarstrand:
## Summary

Added keywords to the most of the packs config.yml files.
Some packs was not found from the Excel file.


polfliet:
Thanks @EmilHammarstrand. The PR looks good to me but I would like someone else to review as well!

packs/roku/config.yml
```diff
@@ -13,3 +13,4 @@ documentation:
     description: Build applications that deliver high-quality video to the Roku platform.
     url: >-
       https://docs.newrelic.com/docs/agents/open-source-licensed-agents/roku-open-source-video-agent
+keywords: .nan
```
aswanson-nr, line 16: And another!

packs/signl4/config.yml
```diff
@@ -18,3 +18,4 @@ documentation:
       Mobile alerting SaaS that bridges the last mile from IT systems, machines,
       and sensors to engineers, IT staff and workers in the field.
     url: https://www.signl4.com/blog/portfolio_item/new_relic_mobile_alerting/
+keywords: .nan
```
aswanson-nr, line 21: Another `.nan`

packs/slack/config.yml
```diff
@@ -19,3 +19,4 @@ documentation:
       by American software company Slack Technologies.
     url: >-
       https://docs.newrelic.com/docs/alerts/new-relic-alerts/managing-notification-channels/notification-channels-control-where-send-alerts
+keywords: .nan
```
aswanson-nr, line 22: One more!

packs/victorops/config.yml
```diff
@@ -19,3 +19,4 @@ documentation:
       closed. Notifications can include charts about the incident.
     url: >-
       https://docs.newrelic.com/docs/alerts/new-relic-alerts/managing-notification-channels/notification-channels-control-where-send-alerts
+keywords: .nan
```
aswanson-nr, line 22: Another!

packs/xmatters/config.yml
```diff
@@ -15,3 +15,4 @@ documentation:
       issues.
     url: >-
       https://docs.newrelic.com/docs/alerts/new-relic-alerts/managing-notification-channels/notification-channels-control-where-send-alerts
+keywords: .nan
```
aswanson-nr, line 18: Last one!

packs/android/config.yml
```diff
@@ -19,3 +19,4 @@ documentation:
       source software.
     url: >-
       https://docs.newrelic.com/docs/mobile-monitoring/new-relic-mobile-android/get-started/introduction-new-relic-mobile-android
+keywords: mobile, mobile agent
```
aswanson-nr, line 22: YAML will treat this syntax as a string instead of a list. To make it a list/array, we will need to update it to this syntax
```yaml
keywords:
  - mobile
  - mobile agent
```

aswanson-nr (changes requested):
This looks fantastic!! I just noticed a `.nan` value in a few spots and that the syntax for the `keywords` field needs to be updated to be a list of strings instead of a single string.

packs/android/config.yml
nr-kkenney: if its easier, wrapping the values in brackets is also equivalent.

```yaml
keywords: [mobile, mobile agent]
```

aswanson-nr:
I resolved all my comments about `.nan` since @rudouglas already found them!

polfliet:
@EmilHammarstrand can you update the script when in case there are no keywords defined in the sheet, it does not add the keywords field, because that currently gives the `.nan`. @rudouglas @aswanson-nr you good with that approach or would you suggest a different approach? (We definitely don't want the `.nan` to show up.)

About the format, the original spec was a comma delimited string, as implemented by Emil. @kidk, has the spec changed to accept an array? I want to make sure this is aligned with how the API is currently implemented before Emil makes that change.

Thanks all for chiming in!

rudouglas:
ye i don't see any issues with that approach at all @polfliet

EmilHammarstrand:
I have made a new commit to this PR where I updated the keywords syntax as a list instead of a string.

aswanson-nr (approved):
Looks good to me!

aswanson-nr:
@EmilHammarstrand Could you go in and sign the CLA? We will need that before merging. If you already did and it's still yelling at you, it's probably because the commits are signed with a different email address than your github account. To fix this, you just need to click on the link in the CLA comment to add the email to your account.

EmilHammarstrand:
Weird, I have signed the CLA and the email in the commits are the same as the email in my github account but still not working 🤔

polfliet:
@EmilHammarstrand is going to try to create a new PR because we don't see the problem with the CLA, he has accepted it and the email address from the commit seems to be matching his GH user. If the other PR works, we will close this one

polfliet:
Closing this PR since the CLA check is now successfull in the new PR: https://github.com/newrelic/newrelic-observability-packs/pull/204

PR #220, opened 2021-09-10
feat(install): add install plans

jbeveland27:
1. Updated the template to include `installPlans` in `_template/config.yml` and removed instrumentation folders

2. Adds install plan definitions for:

    - 8 APM agents
    - Guided Install
    - 16 supported OHIs (Targeted Install)
    - Public API Performance 

3. Added the `installPlans` refs in these Quickstarts:

    - The 16 OHIs
    - The 8 APM agents
    - The 6 Infra host Quickstarts (just the basic linux host OS Quickstart)

----

Will add support for docker/k8s/mac guided install flows in another PR.


install/apm/node/install.yml
```diff
@@ -0,0 +1,26 @@
...
+target:
+  type: agent
+  destination: application
+  os:
+    - linux
+  context: |
+    ## WARNING
+    Using the command will automatically stop and attempt to restart all instrumentable applications, with all the same context, during the installation process. Once instrumented, redeploying any application(s) will require you to re-run the command. If you would like to manually install, please see our Node standard installation.
+
+    ## INFO
```
kidk, line 15: Should we change the `INFO` and `WARNING` to `Info` and `Warning`?

kidk: Everything else looks great, thanks @jbeveland27

jbeveland27: Yeah good callout. So currently I just copied what we display in the existing UI. I'd like to treat think this `context` field like markdown, and then the UI can parse and render that out in the sidebar. However there are some details that aren't fully figured out yet with limitations that might impose (such as preferred styling). I'll leave it like this for now, and we can revisit when we start to implement this in the UI.

image

PR #234, opened 2021-09-14
Move generate-uuid job under the release workflow

### LINKED ISSUE #210: `generate-uuid` workflow failing for forks
## Description
The workflow that currently generates UUIDs for quickstarts is triggered by the `pull_request` event. When this event is triggered from a forked PR, write permission is not given. The `pull_request_target` event helps with this issue by allowing writes from a forked PR, but using that event can allow for arbitrary code execution if the fork's content is checked out and ran. 
To work around this problem Github suggests that you have a `pull_request` triggered workflow that runs all the unsafe code (since it doesn't have write permissions) and saves its output to an artifact. Then have another workflow that is triggered by the completion of the previous workflow, that goes in and takes the privileged action. 

This issue with using that recommendation for our workflow is that we need access to the fork's changes because we want to add a new commit. I don't see an obvious way to determine which quickstarts need UUIDs, generate them, and then commit them in that suggested model.

### Possible solution
One idea I've had is to move this workflow to something that only runs on commits to the `main` branch and commits directly there. Although there would be some concerns around generating extra github releases on each commit and possibly triggering an infinite loop of commits. 

### Reference material
* [pull_request_target event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target)
* [how to not get pwned by actions](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

### Acceptance Criteria 

- [ ] Decide how this should work in the timebox
- [ ] Implement the approach (create a new ticket)

aswanson-nr:
## Summary

An idea for moving the `generate-uuid` job under the release workflow.

## Related issues
- Closes #210


.github/workflows/postmerge.yml
```diff
@@ -0,0 +1,176 @@
...
+        id: generate-uuids
+        run: cd utils && yarn generate-uuids
+
+      - name: Commit changes
+        id: commit-changes
+        run: |
+          git config --local user.email "${{ env.BOT_EMAIL }}"
+          git config --local user.name "${{ env.BOT_NAME }}"
+          git add ./packs/*
+          git diff-index --quiet HEAD ./packs/* || git commit -m 'chore: generate UUID(s)'
```
aswanson-nr, line 40: We will most likely need to add a `--skip-ci` flag to this to prevent an infinite loop of workflows

nr-kkenney (approved):
lgtm. a small comment i have but still would approve as is, is the name of the workflow. i think keeping it as release is my slight preference since that describes what is running (the release process), versus when its running (post-merge) which i think is more valuable as a name.

nr-kkenney (approved):

PR #251, opened 2021-09-15
feat: Add dashboards, alerts, descriptions for Java related quickstarts

polfliet:
Upleveling all Java related quickstarts by:
* Adding the dashboard and alerts from the Java pack (note that we are now duplicating them, because we don't yet have a reference system implemented)
* Adding a default description that speaks more to the value of the quickstart and what users should expect

Thanks @bradleycamacho for providing the copy!


packs/java/adobe-cq/alerts/HighCpuUtilization.yml
```diff
@@ -0,0 +1,29 @@
+---
+name: High CPU Utilization
+
+details: |+
+  This alert is triggered when the CPU Utilization is above 225.
```
kidk, line 5: Why 225? Magic number?

kidk (approved):

packs/java/adobe-cq/alerts/HighCpuUtilization.yml
polfliet: updated!

kidk (approved):

PR #264, opened 2021-09-17
Added install plans to more quickstarts

kidk:
Added a couple of install plans to partner quickstarts, removed a legacy quickstart, and other small clean up.


kidk:
@jbeveland27 Can you double check my install plans?

install/on-host-integration/kafka/install.yml
```diff
@@ -0,0 +1,14 @@
...
+title: Kafka Monitoring Integration
+description: |
+  The New Relic Kafka on-host integration reports metrics and configuration data from your Kafka service. We instrument all the key elements of your cluster, including brokers (both ZooKeeper and Bootstrap), producers, consumers, and topics.
+
+target:
+  type: on-host-integration
+  destination: host
+
+install:
+  mode: link
```
polfliet, line 12: do we have a targetedInstall for Kafka?

polfliet (dismissed):

install/browser/browser-docs/install.yml
```diff
@@ -0,0 +1,14 @@
+id: browser-docs
+name: Browser
+title: Browser Monitoring
+description: |
+  With New Relic One's browser monitoring solution, you get full visibility into the complete webpage life cycle of your application or website. Browser measures page load timing, also known as Real User Monitoring (RUM).
+
+target:
+  type: integration
+  destination: cloud
```
jbeveland27, line 9: Is browser technically an agent? I'd think to make this:

```suggestion
  type: agent
  destination: application
```

Maybe that's wrong though ¯\_(ツ)_/¯

install/on-host-integration/kafka/install.yml
jbeveland27: Unfortunately, no 😕 We set out to have one initially but the setup required a lot of variables set ahead of time, and we never progressed to adding it in.

jbeveland27 (commented):
Left a comment

jbeveland27 (approved):

polfliet (approved):

Issue #149, opened 2021-08-18
[Quickstart Repo] Update repo documentation for quickstarts


jpvajda:
We want to ensure we don't try to document things in many places in this repo. the main source of truth for building a quickstart should be 

1. The [main json config file](https://github.com/newrelic/newrelic-observability-packs/blob/main/utils/schemas/main_config.json)

We don't need these files at go live

- [dashboard config](https://github.com/newrelic/newrelic-observability-packs/blob/main/docs/dashboard-spec.md)
- [alert config](https://github.com/newrelic/newrelic-observability-packs/blob/main/docs/alert-spec.md)
- [pack config](https://github.com/newrelic/newrelic-observability-packs/blob/main/docs/config-spec.md)
- [install config](https://github.com/newrelic/newrelic-observability-packs/blob/main/docs/install-spec.md)

### Acceptance Criteria

- [x] Confirm approach with @jbeveland27 (see comments below)
- [x] Update [main json config file](https://github.com/newrelic/newrelic-observability-packs/blob/main/utils/schemas/main_config.json)
- [x] Once that is in place deprecate those config files as they were "specs" for us to work against during the development of the project
- [x] Before spec files removing let the virtuoso team we are cleaning this up.
- [x] ensure readme and contributing doc don't have documentation redundancies

aswanson-nr:
I'm not sure we want to point users at graphical to look over documentation, since that API schema will be less strict than the JSON schema within the repository. It's possible to generate HTML or markdown from the schemas and host it on github pages or within one of our sites.
Here is one doc generator I've found. https://pypi.org/project/json-schema-for-humans/

jpvajda:
some comments from @jbeveland27  to consider 

> If you're updating the JSON schema automatically, I'd assume that means you're pulling that definition from our NerdGraph API, which isn't an accurate representation of what the source config files allow. The service adds fields on top of what's there. And as I see it, modifying the agreed-upon spec definition should be a bit of a manual process between teams as this is the API contract we've designed. (*However I reserve the right to be wrong about all assumptions of the above )

Nice developer guides/contributor guides that spell things out clearly is :fire:  and I completely support that :100: I just think we need to be more particular with our spec files because these represent an API boundary between teams. 

The updates in that PR show that as an example: there are fields the service is adding explicitly that contributors shouldn't add. Encoding that in the JSON schema isn't the issue, it's that they don't really belong because they're an implementation detail that users need not know/care about

[...]

PR #267, opened 2021-09-17
Simplify documentation

nr-kkenney:
## Summary

* Updated `docs` directory to have human-readable markdown versions of the json schemas
* Added workflow to generate documentation automatically
* Minor touchups to other documentation

Closes: #149


docs/alert_config.md
```diff
@@ -0,0 +1,62 @@
+# Observability Pack Alert Configuration
+
+**Title:** Observability Pack Alert Configuration
+
+| Type                      | `object`                                                                  |
+| ------------------------- | ------------------------------------------------------------------------- |
+| **Additional properties** | [[Any type: allowed]](# "Additional Properties of any type are allowed.") |
+|                           |                                                                           |
```
zstix, line 8: Looks like our script is adding an extra row at the bottom of these tables.

zstix (commented):
Lookin' good! One minor comment about extra table rows getting generated. I'm not sure if there's a setting we can utilize to help prevent that.

🐍

docs/alert_config.md
nr-kkenney: i double checked through their documentation & don't see an option to disable that :rip:

zstix (approved):

PR #276, opened 2021-09-18
chore: fixed several quickstarts

jpvajda:
This PR fixes some  misc issues quick starts I saw

1. SVG logo was too small
2. browser example had no summary and was formatted oddly
3. trend micro had a name in the DB config with no spacing


packs/trendmicro-cloudone-conformity/dashboards/trendmicro-cloudone-conformity.json
```diff
@@ -154,9 +154,7 @@
               }
             ]
           },
-          "linkedEntityGuids": [
-            "MjcyNjYwNXxWSVp8REFTSEJPQVJEfDE1MTgyMDI"
-          ]
+          "linkedEntityGuids": ["MjcyNjYwNXxWSVp8REFTSEJPQVJEfDE1MTgyMDI"]
```
kidk, line 157: I wonder if these will work if they are imported into other accounts. cc @polfliet

kidk (changes requested):
One open question on the linkedEntityGuids. Everything else looks good 👍

packs/trendmicro-cloudone-conformity/dashboards/trendmicro-cloudone-conformity.json
polfliet: Here is the response from the Dashboard team:

> Dashcow doesn’t process the linkedEntityGuids it receives, it just stores them in the database. So if another customer imports that dashboards and they don’t have access to MjcyNjYwNXxWSVp8REFTSEJPQVJEfDE1MTgyMDI, then they wouldn’t be able to retrieve the entity outline (name, etc) of that linked dashboard as Nataly pointed out, and facet linking won’t work.

So if I understand correctly the user would still see the dashboard and that specific widget, but would not be able to use the advanced features like facet linking (which is ok I think)

jpvajda: @polfliet ok oh, I'm not super familiar with how this works but it sounds like this is 🆗

polfliet (approved):

PR #278, opened 2021-09-19
Cleaned up infra requirement

kidk:
Removed the infra requirement on targeted install OHI's + added the requirement on doc installs.


jbeveland27 (approved):
LGTM :shipit: 

As a follow up - we should make sure the actual `install.yml` for all the OHIs have some sort of description that indicates the infra agent will be installed as well

kidk:
Agreed Justin. I need to double check how the description looks in the UI and add a note.

PR #280, opened 2021-09-20
Added notice on OHI's that we will install New Relic Infra

kidk:
We removed the infra step so users won't know we need to install New Relic Infra. This adds a notice in the UI that we will add it. 

Before we merge this I do have a couple of questions.

### DIFF
--- a/install/on-host-integration/apache/install.yml
+++ b/install/on-host-integration/apache/install.yml
@@ -4,6 +4,8 @@ title: Apache monitoring integration
 description: |
   Our Apache integration sends performance metrics and inventory data from your Apache web server to the New Relic platform. You can view pre-built dashboards of your Apache metric data, create alert policies, and create your own custom queries and charts.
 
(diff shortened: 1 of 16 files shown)


kidk:
@jbeveland27 Do you know if we can use Markdown in the description?

jbeveland27:
@kidk From an API standpoint, markdown will work fine. However the UI isn't doing anything to parse and render markdown for the `description` field at the moment (from what I know), so that'd have to be added

kidk:
Thanks @jbeveland27 Let's leave it as is for now, and improve after launch.

jbeveland27 (approved):

jbeveland27:
@kidk I'll merge and redeploy the service to pull these changes in. I think you might want to double check to see if the content lines up the way you'd expect. I don't think the newlines will actually translate to anything, so will have to see how it renders.

PR #284, opened 2021-09-20
Ru/importer renaming

rudouglas:


importer-config.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
 # How to find your accountId: https://docs.newrelic.com/docs/accounts/accounts-billing/account-setup/account-id/
-accountId=
+accountId=10956800
 # How to create a User API Key: https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/#user-key-create
-apiKey=
+apiKey=NRII-5DYAYH7YpiNVIThYHXRz75z1Ld8ls_9N
```
jbeveland27, line 6: @rudouglas You're gonna want to rotate this API key 😄

.github/actions/import-data/src/importer.ts
```diff
@@ -86,10 +86,10 @@ const createPolicy = async (accountId: number, pack: string) => {
 	return response.alertsPolicyCreate.id;
 };
 
-const createDashboardLocal = async (accountId: number, pack: string) => {
+const createDashboardLocal = async (accountId: number, quickstart: string) => {
 	const replacer = new RegExp('"accountId":0', 'g');
 
-	const dir = `${__dirname}/../../../../packs/${pack}/dashboards`;
+	const dir = `${__dirname}/../../../../quickstarts/${quickstart}/dashboards`;
```
jbeveland27, line 92: One thing that stands out here - quickstarts can now be nested, so this likely doesn't work unless the user realizes `path/to/quickstart` is how the input would need to be structured

jbeveland27 (approved):
Seems like this should work overall (or at least as much as it "worked" previously) 😄

importer-config.sh
rudouglas, line 6:
```suggestion
apiKey=
```

rudouglas: says you 😂

importer-config.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
 # How to find your accountId: https://docs.newrelic.com/docs/accounts/accounts-billing/account-setup/account-id/
-accountId=
+accountId=10956800
```
rudouglas, line 4:
```suggestion
accountId=
```

.github/actions/import-data/src/importer.ts
rudouglas: ya good point, i think we can follow up on this post GA

PR #289, opened 2021-09-20
Add messaging/labeling around new and stale PRs/issues

LizBaker:
Closes https://github.com/newrelic/newrelic-observability-packs/issues/218

### DIFF
--- a/.github/workflows/stale-message.yml
+++ b/.github/workflows/stale-message.yml
@@ -13,9 +13,9 @@ jobs:
     steps:
       - uses: actions/stale@v4
         with:
-          stale-issue-message: 'Old issues will be closed after 105 days of inactivity. This issue has been quiet for 90 days and is being marked as stale. Reply here or add the label "not-stale" to keep this issue open.'
+          stale-issue-message: "Old issues will be closed after 105 days of inactivity. This issue has been quiet for 90 days and is being marked as stale. Reply here to keep this issue open."
           close-issue-message: "This issue is being closed due to inactivity. Is this a mistake? Please re-open this issue or create a new one."
-          stale-pr-message: 'Old PRs will be closed after 30 days of inactivity. This PR has been quiet for 14 days and is being marked as stale. Reply here or add the label "not-stale" to keep this PR open.'
+          stale-pr-message: "Old PRs will be closed after 30 days of inactivity. This PR has been quiet for 14 days and is being marked as stale. Reply here to keep this PR open."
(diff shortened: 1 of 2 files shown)


.github/workflows/stale-message.yml
```diff
@@ -0,0 +1,29 @@
...
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
+          stale-issue-message: 'Old issues will be closed after 105 days of inactivity. This issue has been quiet for 90 days and is being marked as stale. Reply here or add the label "not-stale" to keep this issue open.'
```
jpvajda, line 16: I don't know if all user can add labels, so we may just ask for a comment here.

.github/workflows/stale-message.yml
```diff
@@ -0,0 +1,29 @@
...
+          close-issue-message: "This issue is being closed due to inactivity. Is this a mistake? Please re-open this issue or create a new one."
+          stale-pr-message: 'Old PRs will be closed after 30 days of inactivity. This PR has been quiet for 14 days and is being marked as stale. Reply here or add the label "not-stale" to keep this PR open.'
```
jpvajda, line 18: Same comment on asking users to add a label.

jpvajda (changes requested):

.github/workflows/stale-message.yml
LizBaker: i feel like the user that initiated the PR can usually add labels but i'll find out for sure and if not we can get rid of that suggestion

LizBaker: hmm didn't actually find any definitive answer about PRs but i did see some for labels not being available. we do say _or_  add a label, do you think that's clear enough or would you rather we completely remove it?

.github/workflows/stale-message.yml
aswanson-nr: The stale bot automatically handles adding and removing the labels, all the user needs to do is update the PR/issue.

.github/workflows/stale-message.yml
LizBaker: yeah we had just written in the message that you could either comment or add a not stale label to de-stale the issue/PR. but we can change the language to not tell folks to add a label

aswanson-nr (dismissed):
LGTM

aswanson-nr (approved):

jpvajda (commented):
LGTM

PR #292, opened 2021-09-21
Level updates

polfliet:
Working with GTS on updating all support levels


jpvajda (dismissed):
Since so many files have changed I assumed these are all valid changes, so 👍  if we are converting this many quickstarts to New Relic.

Does this mean all of these will be supported by GTS now?

PR #300, opened 2021-09-21
chore: create Infrastructure quickstart description

aliandersonsn:
Search Nurture's Infrastructure recommendations

## Summary

A concise description of the changes being introduced.
If this is a new Observability Pack, please provide a description of its use cases and a list of the Pack's components.

### Screenshots

Images of any visual changes, such as a dashboard.


packs/infrastructure/config.yml
```diff
@@ -8,7 +8,43 @@ title: Infrastructure
 
 # Description of the pack
 description: |
-  The infrastructure pack allows you to get visibilility into the performance and availability of your hosts and operating system.
+  ## Infrastructure Monitoring Quickstart
```
jpvajda, line 11:
```suggestion
  ### Infrastructure Monitoring Quickstart
```

jpvajda: Making a H3 per our discussions on not using an H2.

packs/infrastructure/config.yml
```diff
@@ -8,7 +8,43 @@ title: Infrastructure
...
+
+  Infrastructure monitoring provides deep visibility into performance, available resources, and your entire estate at a glance. Remote infrastructure monitoring tools collect and display data and related attributes represented by various metrics and metadata.
+
+  Monitoring helps cloud-based DevOps teams quickly identify and resolve issues, averting potential downtime. By triggering alerts, server monitoring also improves overall flexibility and scalability to manage peaks. You can drill down into specific errors and resolve them to ensure that your environment is fully optimized.
+ 
+  New Relic's infrastructure quickstart provides visibility across the entire stack and enables greater scale and efficiency. It's a monitoring solution for multifaceted hybrid environments with customized dashboards and visualizations.
+
+
+  ### What's Included?
```
jpvajda, line 20: Having what's included twice seems odd to me. The tiles above the description tells you what's included, so i'd suggest we don't use this pattern or if it's important call it something else.

jpvajda: Screen Shot 2021-09-23 at 12 45 02 PM

packs/infrastructure/config.yml
```diff
@@ -8,7 +8,43 @@ title: Infrastructure
...
+  #### New Relic Infrastructure Monitoring
+
+  Instant observability with New Relic quickstarts minimizes the complexity in efficiently managing enterprise infrastructure. You can see everything at a glance with New Relic's infrastructure dashboard.
+
+  The dashboards in this quickstart enable DevOps teams to quickly see key components to their infrastructure. System load dashboards give key insights into how heavy the load is on your server. It can help prevent loss of full functionality from too many system processes at a given time. 
+
+  New Relic's instant observability quickstart helps DevOps engineers reduce complexity and enhance efficiency through unmatched visibility in an infrastructure dashboard. 
+
+
+  ### What Makes New Relic Unique?
```
jpvajda, line 41: Not sure we need a pitch for new relic in individual quickstarts, it seems unnecessary and adds to the length of the markdown which we want to try to be more succinct.

jpvajda (changes requested):

packs/infrastructure/config.yml
bradleycamacho: I think we should include it, as the dashboard is much harder to "skim" than the text here is. This gives the user the specifics of what is included in the pack.

I do agree that having "What's included" repeated twice is bad practice though. I agree we should change this instance , I'll push up a commit with a change.

packs/infrastructure/config.yml
bradleycamacho: I would argue we should make the pitch. My understanding is that one of the goals for I/O is to drive new user sign-ups. Assuming portion of our users will have their first experience with New Relic through quickstarts, I think it's important to highlight the value of the entire platform.

aswanson-nr:
The repository has been renamed to `newrelic/newrelic-quickstarts`. Github will handle routing your remotes to this new url, but we recommend that you update your remote as soon as possible.
**NOTE**: As part of this change, we have moved all the quickstarts from the `/packs` directory to the `/quickstarts` directory. To bring your pull request up to date, please pull in the latest commits from the `main` branch.

Use the following command to set your new remote:
```bash
git remote set-url origin https://github.com/newrelic/newrelic-quickstarts.git
```

Or with an SSH origin
```bash
git remote set-url origin git@github.com:newrelic/newrelic-quickstarts.git
```

packs/infrastructure/config.yml
bradleycamacho: I've pushed up a change resolving this.

bradleycamacho (approved):

bradleycamacho (approved):

polfliet:
Closing, these changes were merged as part of https://github.com/newrelic/newrelic-quickstarts/pull/357

PR #302, opened 2021-09-21
chore: minor updates

jpvajda:
provides some minor updates to contributor, readme and pack which had typo in it


packs/lamp/config.yml
```diff
@@ -11,7 +11,7 @@ description: |+
   This LAMP observability pack allows you to monitor a LAMP stack (Linux, Apache, MySQL, PHP).
   It includes a specific dashboard to get observability for an entire LAMP stack.
 
-  NOTE: Observability pack references are currently not yet implemented,
+  Observability pack references are currently not yet implemented,
```
polfliet, line 14: Also update this to quickstarts? or is that part of another PR to move the name?

polfliet (approved):
Just 1 small comment if this PR should also update all `observability pack` references to become `quickstart`

packs/lamp/config.yml
jpvajda: I fixed this and also removed the line about installing other quickstarts as I think this is covered now.

polfliet (approved):

PR #311, opened 2021-09-22
Video and Ads packs

asllop:
## Summary

Added the following packs:

- iOS / tvOS video agent
- Android video agent
- Web video agent, with the following video trackers:
  - JWPlayer tracker
  - HTML5 player tracker
  - ThePlatform player tracker
  - VideoJS player tracker
  - Akamai player tracker
  - Chromecast tracker (not a web player, but dependant on the web video agent because is JS)
- Also depending on the web video agent, the following Ads trackers:
  - Google Publisher Tag (GPT)
  - Prebid


packs/ads/web/ads-web-gpt/config.yml
```diff
@@ -0,0 +1,19 @@
+name: ads-web-gpt
+title: Agent for Google Publisher Tags
+description: Agent to monitor web applications using GPT.
+summary: Agent to monitor web applications using GPT.
+level: Community
+authors:
+  - New Relic
```
polfliet, line 7: We have been using New Relic Labs for contributions from NR Labs. Can you update the author?

polfliet (changes requested):
@jbeveland27 maybe a stupid question but do we support the directory structure as used here packs/ads/web/ads-web-gpt/? We have been using 2 levels deep, but this is 3 and I just want to confirm the API will find those

Thanks @asllop great to have these represented in New Relic I/O!

packs/ads/web/ads-web-gpt/config.yml
asllop: Done!

jbeveland27:
> @jbeveland27 maybe a stupid question but do we support the directory structure as used here packs/ads/web/ads-web-gpt/? We have been using 2 levels deep, but this is 3 and I just want to confirm the API will find those

@polfliet Yep, that'll work. [Here's an example](https://regex101.com/r/OIxMfj/3) of the regex we use. As long as it's in the `packs` directory and there's a `config.yml|yaml` at the end, we'll match

polfliet (approved):

polfliet:
> > @jbeveland27 maybe a stupid question but do we support the directory structure as used here packs/ads/web/ads-web-gpt/? We have been using 2 levels deep, but this is 3 and I just want to confirm the API will find those
> 
> @polfliet Yep, that'll work. [Here's an example](https://regex101.com/r/OIxMfj/3) of the regex we use. As long as it's in the `packs` directory and there's a `config.yml|yaml` at the end, we'll match

thanks for confirming!

Issue #324, opened 2021-09-22
null description values in dashboard json fails schema check


jpvajda:
### Issue 

Copying a dashboard json creates a null value in description which requires human interaction [See PR](https://github.com/newrelic/newrelic-observability-packs/pull/309/commits/1d41dda3637bda17f6d3daaa8280775ec02042fd#diff-6ae35f1370[%E2%80%A6]d98143f927bf05a) could we either convert this to string, or not fail on null? This triggers a failure of the [schema check](https://github.com/newrelic/newrelic-observability-packs/pull/309/checks?check_run_id=3675254586)

### Acceptance criteria

- [ ] could we make better suggestions on what the issue is, and ensure they understand a null value can't be present.

jpvajda:
relates > https://github.com/newrelic/newrelic-observability-packs/issues/242

PR #328, opened 2021-09-22
improve validation output (for null description of dashboard)

nr-kkenney:
## Summary

* maintain previous functionality, but add in new functionality to add
instancePath & schemaPath to messages by default. This context will hopefully
point users to specific problematic fields and checks.
* removed redundant condition in schema json.

## Examples

Previous output:
```sh
Error: packs/mlops/algorithmia/dashboards/algorithmia.json
	 must be string
	 must NOT be valid
```

Now:
```sh
Error: packs/mlops/algorithmia/dashboards/algorithmia.json
	 '/description' must be string
```

## Links
Resolves: #324

### DIFF
--- a/utils/__tests__/validate_packs.test.js
+++ b/utils/__tests__/validate_packs.test.js
@@ -182,6 +182,16 @@ describe('test validateFile', () => {
 
       expect(errors.length).toBe(1);
     });
+
+    test('ensure error output is correct for null description', () => {
+      const dashboardTestFile = getTestFile('dashboard');
+      dashboardTestFile.contents[0].description = null;
+
+      const { errors } = validateFile(dashboardTestFile);
+
+      expect(errors.length).toBe(1);
+      expect(errors[0].message).toBe("'/description' must be string");
+    });
   });
 
   describe('flex validation', () => {
--- a/utils/schemas/dashboard_config.json
+++ b/utils/schemas/dashboard_config.json
@@ -27,12 +27,7 @@
       "default": "",
       "examples": [
         ""
-      ],
-      "not": {
-        "enum": [
-          null
-        ]
-      }
+      ]
     },
     "pages": {
       "$id": "#/properties/pages",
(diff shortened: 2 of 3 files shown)


LizBaker (approved):

utils/__tests__/validate_packs.test.js
```diff
@@ -182,6 +182,16 @@ describe('test validateFile', () => {
...
     });
+
+    test('ensure error output is correct for null description', () => {
+      const dashboardTestFile = getTestFile('dashboard');
+      dashboardTestFile.contents[0].description = null;
+
+      const { errors } = validateFile(dashboardTestFile);
+
+      expect(errors.length).toBe(1);
+      expect(errors[0].message).toBe("'/description' must be string");
```
jpvajda, line 193: would it be overkill to clearly state this is the dashboard description just to make it a little easier to resolve?

jpvajda (changes requested):

utils/__tests__/validate_packs.test.js
nr-kkenney: the error does point to the specific file that failed, so that context should indicate its a dashboard.

if we wanted the word dashboard for this one field, that might be awkward, but is doable.

jpvajda: That's cool. thanks for explaining!

jpvajda (approved):

polfliet:
Great to give users more info about where the problems occur!
One question: `description = null` is how the dashboards are exported from the product. I think we should avoid putting toil on a contributor where they are exporting a dashboard and then have to manually go replace the null values by "". How could we improve that?

nr-kkenney:
@spolfliet great question. ill try my best to answer for what i know, and point out what i dont know.

to start, i think this is completely doable in a number of ways, but we may consider it awkward.

i think the thing we would have to address -- assuming this has to be a non-null string at some point -- is where the conversion happens, and can we pass null along the way.

i think it would be easy to have null be a valid type, value for the repository, but then we would either need to convert it to string for the consumer that reads the repository, or make sure null is an acceptable value for them (which may require a code change). i dont think we want to convert it at the repo level, since the automation would be complicated (more complicated anyways than if it was within the api / consumer) and it conceptually conflicts with the validation we are doing -- you can submit an invalid dashboard, but we will make it valid.

PR #373, opened 2021-09-27
feat: Add featured keyword to partner quickstarts

polfliet:
Adding the `featured` keyword to the launch partner related quickstarts


jpvajda:
Nice.

PR #378, opened 2021-09-27
fix: build importer and update readme

rudouglas:


aswanson-nr (approved):
Would love to get approval from @jbeveland27 as well!

jbeveland27 (approved):
:shipit:

PR #400, opened 2021-09-29
chore: update workflow to move prs to projeect

rudouglas:
If we add the `eng` label this workflow won't run

### DIFF
--- a/.github/workflows/pr-open.yml
+++ b/.github/workflows/pr-open.yml
@@ -16,7 +16,17 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      - name: Comment and label
+      - name: Comment, label and move to project board
         run: |
           gh pr comment ${{github.event.pull_request.number}} --body "Thank you for your contribution, our team will be reviewing this shortly, please be available for any follow up questions or code review feedback!"
           gh pr edit ${{github.event.pull_request.number}} --add-label "triage-needed"
+
+  automate-project-columns:
+    if: "!contains( github.event.pull_request.labels.*.name, 'eng')"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Move to Project Board
+          gh pr edit ${{github.event.pull_request.number}} --add-project "NR1 Quickstarts Community"


LizBaker:
@rudouglas i might be blind but which part determines that this doesn't run if we have an eng label?

rudouglas:
Oh whoops sorry @LizBaker that was the initial pass at this, JV wanted it to only route to the Project Board if all changed files were in the `quickstarts` folder, so only content updates.

.github/workflows/pr-open.yml
```diff
@@ -16,7 +16,55 @@ jobs:
...
           gh pr edit ${{github.event.pull_request.number}} --add-label "triage-needed"
+
+  check_files:
+    name: Check files
+    outputs:
+      run_job: ${{ steps.check_files.outputs.move_to_project }}
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout code
```
aswanson-nr, line 31: Unfortunately, this checkout leaves us open to the type of attacks that can steal credentials. I get wanting to filter out PRs that don't change quickstarts, but I think the little bit of toil needed to remove extra PRs from the project is better than the alternative. @jpvajda Would you be fine with all PRs routing to that project?

aswanson-nr (changes requested):
I'm a little worried the `check_files` step can leave us open to GH `pwn` requests.

.github/workflows/pr-open.yml
nr-kkenney: i agree with the approach of just including everything. however, if we really wanted to filter out files, i dont think we need to check anything out. 

similar to the translation workflow, we can just hit the public git api to see what files are changed in the pr. since the repo is public, we should be able to use a read only token / trash token to get results.

aswanson-nr: That's a nice way to do it as well! I'd be totally fine with going that direction. The only reason I suggested just routing everything to the board is that is how the docs team handles it and it seems to work fine for them. @jpvajda  Would love to get your input on this.

jpvajda: ideally we just show quickstart contributions on this board, not every PR in the repo, I think by showing all  PRS, we sorta defeat the purpose of the board, as it's supposed to aid someone in being able to quickly see all the related quickstarts activity of the project and not have to filter through all the noise of every PR

aswanson-nr: Okay cool, then @nr-kkenney's approach should get us what we need

jpvajda: 👍

rudouglas: @nr-kkenney will that still allow us to route the project to our board, can we use a read-only token for that as well?

nr-kkenney: good call. i totally skipped over that in my mind 🙈 . would need write permissions for that i think.

rudouglas: Ye it looks like the Docs triage uses the same method to move PR's to the board, will that need to be readdressed as well? @aswanson-nr should we default to all PR's to the board for now?

nr-kkenney: @rudouglas that workflow was designed with all this in mind. it doesn't check out code, or operate on anything in the repository.

rudouglas: oh riiiight ok cool i'll just rejig this so it follows that format. My original method for this was to send new PR's to the board if they didn't have the `eng` label on them. The idea there was that when we are creating our internal PR's we could just put that label on so only outside contributor PR's route to the board. That didn't require any checking out of code either so could also be an option

aswanson-nr: @rudouglas You could always move this to a different workflow that only triggers on `pull_request_target` with changes to the `/quickstarts` directory. Then that would allow you to ignore the API or the fork's branch.

nr-kkenney: big 🧠 . right there ^

rudouglas: oh hello 🧠 🦢 might need a lil help on this if we have time to 🏕️ 🔥

jpvajda:
I'm John Vajda and I approve this PR. 🤣  Good work!

PR #411, opened 2021-10-01
Ru/dashboard images

rudouglas:


CONTRIBUTING.md
```diff
@@ -178,9 +179,23 @@ documentation:
...
+#### Images directory
+
+> See the [docs](https://github.com/newrelic/newrelic-quickstarts/blob/main/README.md#getting-started) for more details on `dashboards`
+
+- These images are `optional` and should contain images you want to display within a markdown widget on your Dashboard.
+- File name should be `quickstart_name01`, `quickstart_name02`, etc
+- These images should be stored in the quickstart's images directory. ex: `/quickstart_name01/images`.
+- Must be in `.png`, `.jpg`, `.jpeg` or `.svg` format
+- Each image file must be less than `4MB` in size
+- There should be no more than `6`  dashboard images per dashboard
```
nr-kkenney, line 191: cant believe i didnt ask this before 🙈 . if a user includes two dashboards, doesnt this mean they can supply 12 images?

does the code currently account for this? looks like it might be looking for just 6 total for the quickstart.

jpvajda: @nr-kkenney good question, I think  we set out to set a 6 dashboard images per dashboard, limit originally but if we are seeing multiple dashboards being addded per pack we may want to consider a higher limit if this accounts for all dashboard images.

rudouglas: @nr-kkenney oooh good call yes that would make total sense, i'll alter this to account for the amound of dashboard json files in the folder

CONTRIBUTING.md
```diff
@@ -166,10 +167,10 @@ documentation:
 
 #### Dashboard images (screenshots)
```
aswanson-nr, line 168: We should probably update this heading so people know images and screenshots are different things.
```suggestion
#### Dashboard Screenshots
```
And then have a separate heading for images displayed on the dashboard

utils/validate_images.js
```diff
@@ -27,31 +28,52 @@ const validateImageCounts = (quickstartDirs) => {
...
+        (p) => p !== iconPath && p !== logoPath && !p.includes(DASHBOARD_IMAGES_PATH)
       );
 
-      if (screenshotPaths.length > MAX_NUM_IMG) {
+      const dashboardImagePaths = imagePaths.filter(
+        (p) => p !== iconPath && p !== logoPath && p.includes(DASHBOARD_IMAGES_PATH)
+      );
+      
+      // Each dashboard is allowed MAX_NUM_IMG dashboards
+      if (screenshotPaths.length > (MAX_NUM_IMG * dashboardCount)) {
```
aswanson-nr, line 55: My understanding is that you can have 6 images included in each dashboard, but you can only have 6 total screenshots for your quickstart. @jpvajda Can you confirm this? Feel free to ignore this if I'm incorrect.

utils/validate_images.js
```diff
@@ -27,31 +28,52 @@ const validateImageCounts = (quickstartDirs) => {
...
+          imageCount: dashboardImagePaths.length,
+          maxImages: MAX_NUM_IMG * dashboardCount,
         };
       }
     })
     .filter(Boolean);
 
   if (directories.length) {
-    core.setFailed('Components should contain less than 6 images');
+    core.setFailed('Components should contain less than 6 images each');
```
aswanson-nr, line 76: We probably want a unique message for screenshots and images, I don't think users would be able to tell if they had a problem with images or screenshots from this message.

utils/validate_images.js
```diff
@@ -27,31 +28,52 @@ const validateImageCounts = (quickstartDirs) => {
...
-
+      
+      // Max images is per dashboard so we need to account for this by getting the number of dashboards
+      const dashboardCount = glob.sync(
+        path.resolve(quickstartDirName, 'dashboards/*.json')
+      ).length;
+      
       const screenshotPaths = imagePaths.filter(
-        (p) => p !== iconPath && p !== logoPath
+        (p) => p !== iconPath && p !== logoPath && !p.includes(DASHBOARD_IMAGES_PATH)
```
aswanson-nr, line 47: Something to consider here, what happens if we have a quickstart named `image` or `images`?

aswanson-nr (commented):
Had a few comments around output and image limits! But overall looks great

utils/validate_images.js
rudouglas: We talked about this and we do specify `6 screenshots` per dashboard for contributions

nr-kkenney (approved):

PR #413, opened 2021-10-01
feat: check for uuid uniqueness

rudouglas:


utils/check_quickstart_uniqueness.js
```diff
@@ -32,30 +32,59 @@ const getMatchingNames = (namesAndPaths) => {
...
+ * @returns {Object[]} an array of matching values
+ */
+ const getMatchingIds = (idsAndPaths) => {
+  return idsAndPaths.reduce((acc, { id, path }) => {
+    const duplicates = idsAndPaths.filter(
+      (quickstart) => quickstart.id && quickstart.id === id && quickstart.path !== path
+    );
+
+    return [...new Set([...acc, ...duplicates])];
+  }, []);
```
nr-kkenney, line 48: i see we copied this from above, but i think i spent a bit of time wondering how this works 🙈  

i think the last line in particular is quite packed with 'magic', that might be worth being unpacked for clarity.

another thing is that this is probably doing more work than needed. the performance doesn't really matter for our usage, but i think we only need to loop over idsAndPaths once (rather than nested) to find duplicates, and at each iteration, we dont have to deduplicate things we've already found.

psedocode:
```none
// map between ids, and associated paths
ids_to_paths_map = { }
foreach id_path in ids_and_paths:
  if new id, add it to map with path
  if already seen, update the map entry for that id with an additional path
```
then we might have something like:
```json
{
  'id_1' : ['path_1', 'path_2']
  'id_2': ['path_3', 'path_4']
}
```

nr-kkenney (approved):
LGTM. i did leave a comment just for personal musing, but not to require a change, just something to think about and chew on.

PR #433, opened 2021-10-06
chore(kentik-firehose): Added featured tag

nr-jbiggley:
# Summary

Added - featured tag as Kentik is a launch partner


jpvajda (approved):
I assume this was approved to be featured so I'l merge this in

PR #549, opened 2021-11-08
Require logo in the schema

rudouglas:


utils/check-missing-logos.js
```diff
@@ -0,0 +1,46 @@
...
+  return noLogos;
+};
+
+
+const main = () => {
+  const quickstartDirs = findMainQuickstartConfigFiles();
+  const quickstartsWithNoLogos = validateLogoExists(quickstartDirs)
+  if (quickstartsWithNoLogos.length > 0) {
+    console.log('<!> Found Quickstarts that do not contain logos');
+    console.log(quickstartsWithNoLogos);
```
nr-kkenney, line 31: i think we want to fail in this case, so we probably want a failure exit code here.

utils/check-missing-logos.js
```diff
@@ -0,0 +1,46 @@
+const {
```
aswanson-nr, line 1: As much as I love a utility script, I think this is only good for a one time use since the repository automation will handle keeping the rest of them schema compliant. If you think this is still worth adding to the repo, let's get a package.json script setup to execute it for ease of use.

aswanson-nr (approved):

aswanson-nr:
Looks like you have a few CI/CD failures with the new schema rules (the unit test failures should be related as well)

utils/check-missing-logos.js
rudouglas: @aswanson-nr Ye the reason i did this was so that the content team could run it to check they had added all the missing logos, they can pull down the PR and work on it from there. Probably also why the checks are failing, they will fail until these are added (except the unit tests i'll fix those). Would it make sense to either 

A: Use a Feature Branch for this or
B: Leave this in draft so the content team can do work using the script

utils/check-missing-logos.js
```diff
@@ -0,0 +1,46 @@
...
+      
+    }).filter(Boolean);
+  return noLogos;
+};
+
+
+const main = () => {
+  const quickstartDirs = findMainQuickstartConfigFiles();
+  const quickstartsWithNoLogos = validateLogoExists(quickstartDirs)
+  if (quickstartsWithNoLogos.length > 0) {
```
zstix, line 29: I think the "quickstarts" qualifier for these variables might be a little too much context. I think we can assume that they're all related to quickstarts in this script.

zstix (approved):
A few small comments regarding code style things.

utils/check-missing-logos.js
```diff
@@ -0,0 +1,46 @@
...
+    .map((quickstart) => {
+      const quickstartConfig = readQuickstartFile(quickstart).contents[0];
+      
+      if (!quickstartConfig.logo) {
+        return quickstart.match(/newrelic-quickstarts(.*)/)[1];
+      }
+      
+      
+    }).filter(Boolean);
+  return noLogos;
```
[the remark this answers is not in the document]
rudouglas: Ye that makes a lot more sense, not sure why i went with the `map` here. We will need to use `config.logo` because some of the configs have `logo: ''` so we need to account for those

utils/check-missing-logos.js
rudouglas: that is a good point ye

jpvajda (approved):
now that all logos are added, this can be merged in

utils/check-missing-logos.js
jpvajda: We went with A, it's done

PR #608, opened 2021-12-02
maint: Update install plans where OS should be windows only or the se…

natecanfield822:
…tup nerdlet should be used

# Summary

-Removing the targetedInstall nerdlet for node and using the setup nerdlet
-Removing targetedInstall nerdlet for php quickstarts that do NOT require php-fpm
  -Created a separate php install plan for php-fpm quickstarts
-Removing linux as a valid OS for dotnet quickstarts that only support windows
  -Created a separate dotnet install plan for windows-only quickstarts


install/apm/node/install.yml
```diff
@@ -8,19 +8,8 @@ target:
...
-    ###### INFO
-    This installation supports instrumenting Node processes managed by PM2. If you are using a different process manager, or are not using one at all, please see our Node standard installation.
-
+    
 install:
-  mode: targetedInstall
-  destination:
-    recipeName: node-agent-installer
-
-fallback:
```
jbeveland27, line 23: this should still have a `fallback` to the docs. The other apm install plans that use a setup nerdlet as the primary install function this way as well. 

Here's the link: https://docs.newrelic.com/docs/apm/agents/nodejs-agent/getting-started/introduction-new-relic-nodejs/#install

install/apm/php/install.yml
```diff
@@ -8,19 +8,8 @@ target:
...
-
-    ###### INFO
-    This single-command will instrument PHP applications on linux systems with php-fpm, nginx, or apache. For other environments, please see our PHP standard installation.
 
 install:
-  mode: targetedInstall
-  destination:
-    recipeName: php-agent-installer
-
-fallback:
```
jbeveland27, line 23: Same thing here - we should have `fallback` to the docs: https://docs.newrelic.com/docs/apm/agents/php-agent/getting-started/introduction-new-relic-php/#installation

install/apm/php-fpm/install.yml
```diff
@@ -0,0 +1,26 @@
+id: php-fpm-agent
```
jbeveland27, line 1: this file looks good, but I'd recommend nesting this under the `php` directory rather than `php-fpm`. And then move the existing php install plan to a new name. Something like this would work:

```
install/apm/php/php-targeted-install/install.yml
install/apm/php/php-setup-nerdlet/install.yml
```

install/apm/dotnet-windows/install.yml
```diff
@@ -0,0 +1,26 @@
+id: dotnet-windows-agent
```
jbeveland27, line 1: Same thing here: I'd move this file and the original `dotnet-agent` like so:

```
install/apm/dotnet/windows/install.yml
install/apm/dotnet/multi-os/install.yml
```

jbeveland27 (changes requested):
Left some comments

install/apm/php/targeted-install/install.yml
```diff
@@ -1,4 +1,4 @@
-id: php-agent
+id: php-fpm-agent
```
jbeveland27, line 1:
```suggestion
id: php-targeted-install
```
Should probably update this id so it matches the directory

quickstarts/php/codeigniter/config.yml
```diff
@@ -16,4 +16,4 @@ keywords:
   - apm
   - php
 installPlans:
-  - php-agent
+  - php-targeted-agent
```
jbeveland27, line 19:
```suggestion
  - php-targeted-install
```
Update this so it matches the `id` of the install plan

jbeveland27 (changes requested):
2 more small changes and I think this is good to go then

jbeveland27 (approved):
🚀

PR #611, opened 2021-12-03
feat: Enable tomcat targeted installs

natecanfield822:
# Summary

Updating the quickstarts for Tomcat and Tomcat EE to used the targeted install with the nerdlet as a fallback.


jbeveland27 (approved):
🍪

PR #614, opened 2021-12-07
Ru/rename quickstarts content

iazam6:
I'm not sure if I did this right, but I have tried to PR multiple quickstarts for the notifications category in I/O. They are all attached below.


jpvajda (changes requested):
Just follow up on my slack message to include this feedback in the PR:

1. I'm not clear on why you created a notifications directory in the quickstarts directory. 

2. Also there are a lot of empty files which we wouldn't want to merge in. So I'm going to request you change this PR to draft  as there seems be some work to do here to get this ready to merge.  
3. I am little confused on why you are creating quicks configuration files independent of an actual quickstart..  
4. finally what is the purpose of the docslink.yml files? Did someone give you direction on taking this approach?  

I think the best way to move this forward is for us to discuss what you are attempting to do with this PR as it seems to breaking a lot of established patterns we have. I'll find some time to talk through it next week.

jpvajda:
After talking to @iazam6 and Matan here is what we decided 

`1` We will close this PR and work from a new one. 
`2`  If you can provide us a list of which quickstarts need to be considered notification types we can get those quickstarts created or updated existing ones in a new PR. 
`3` Given notifications will just be references to docs URLs, just provide us the URLs that we need to reference as well as the copy you'd like added to explain the notification functionality. That can be added to the description markdown. 
`4` We'll get a new PR submitted today and tag you both for a review.  
`5` If we have any questions we can work in this

PR #615, opened 2021-12-07
Atlassian quickstart 

jpvajda:
# Summary

Adds a quickstart for JIRA

## Pre merge checklist

- [x] Did you check your descriptive content for voice, tone, spelling and grammar errors?

### Screenshots

Attach images of any visual changes, such as a dashboard here.


jpvajda:
This still needs an install plan which I will create.

quickstarts/atlassian/atlassian-jira/atlassian-jira.yml
```diff
@@ -0,0 +1,47 @@
+# Name of the quickstart (required)
```
josephgregoryii, line 1: does the `id` get generated automatically? [reference](https://github.com/newrelic/newrelic-quickstarts/blob/main/quickstarts/aws/aws-api-gateway/config.yml#L1)

josephgregoryii (dismissed):

quickstarts/atlassian/atlassian-jira/atlassian-jira.yml
aswanson-nr, line 1: Currently, our system only supports this file being called `config.yml`. You'll need to change the name in order for this to be installable and viewable within the system.

aswanson-nr (changes requested):
Hey Y'all, this looks great! You'll just need to change the name of your main config file from `atlassian-jira.yml` to `config.yml` and things should be good to go!

quickstarts/atlassian/atlassian-jira/atlassian-jira.yml
jpvajda: The `id` of the quickstart will autogenerated, but the `id` of the install plan is user defined

jpvajda: ahhh I knew this.. 🤣  .. good catch. fix coming in hot

jpvajda: This would make a nice automated test.. it be cool to run some tests on PR submission to flag things like this before the PR is submitted.

josephgregoryii (dismissed):

jpvajda:
One more thing to look at 
> The id for the install plan looks off

quickstarts/atlassian/atlassian-jira/config.yml
```diff
@@ -0,0 +1,47 @@
...
+keywords:
+  - JIRA
+  - Atlassian
+  - Notifications
+  - AIOps
+  - Incident Intelligence
+  - Destinations
+
+installPlans:
+  - third-party-atlassian-jira
```
aswanson-nr, line 36: The path is not factored into the id of an install plan, you'll just need to update this:
```suggestion
  - atlassian-jira
```

jpvajda: see my recent commit which sets the install plan to third-party-atlassian-jira, which follows the naming convention for all install.ym files in the third party directory https://github.com/newrelic/newrelic-quickstarts/tree/main/install/third-party

aswanson-nr: Notice that those install plan ids have the string `third-party` prefixed to their ids. The install plan for this quickstart does not have that prefix.

jpvajda: the sendgrid third party follows the pattern I chose to go with where the installPlan ID is  `third-party-atlassian-jira` and the reference to the plan is  `third-party-atlassian-jira`

I didn't check everyone though, but I assume this is the pattern to follow and as long as the `id` matches the the reference to that `plan` it will work.

as a note: this is why I want to document the creation of the install plans better.

jpvajda: it looks like i didn't push... 💯

aswanson-nr (approved):

PR #625, opened 2021-12-09
fix: adding guided-install on Log Analysis

thcanate:
# Summary
change the installPlan of Log analysis dashboard from log-docs to guided-install.


quickstarts/audit/log-ingestion-analysis/config.yml
```diff
@@ -30,7 +30,7 @@ keywords:
 # Reference to install plans located under /install directory
 # Allows us to construct reusable "install plans" and just use their ID in the quickstart config
 installPlans:
-  - log-docs
+  - guided-install
```
jpvajda, line 33: I just want to confirm, by added this you will be invoking this [installPlan ](https://github.com/newrelic/newrelic-quickstarts/blob/main/install/guided-install/default/install.yml)

which on click of install will open the `nr1-install-newrelic.nr1-install-newrelic` nerdlet for the guided install flow in the platform, is that the expectation here?

jpvajda (changes requested):

quickstarts/audit/log-ingestion-analysis/config.yml
thcanate: The idea is to be able to install the infra agent to forward the logs or you can skip the installation. but you still are able to install the dashboard, otherwise if I leave the log-docs your are forwarded to the documentation and the dashboard is not installed.

jpvajda: Perfect! Just wanted to double check.

jpvajda (approved):

PR #648, opened 2021-12-16
Fix/remove outlier alerts

jbeveland27:
This PR is just some cleanup 🧹 

The ability to create/update outlier alerts has been deprecated in NerdGraph, so we need to remove all references of outlier alerts from quickstarts (since the underlying API is no longer supported). Fortunately `outlier` alert conditions were only really present in templates, the jumpstart quickstart, and in the schema validation.


jbeveland27:
Note: the `import-data` github action also supported outlier alerts, so I removed it from there as well just in case anyone still uses that.

jpvajda:
I think to support this change we'll also want to remove `outlier` from this code https://github.com/newrelic/developer-website/search?q=outlier

rudouglas (approved):
Superb deletions

PR #653, opened 2021-12-20
chore: fixes casing on notifications

jpvajda:
fixes the casing on the notifications, it should be lower case.


zstix (approved):
Kinda weird that this is supposed to be lower case, but all the other keywords in this list are capital case. Is this documented anywhere?

PR #660, opened 2021-12-22
[Repo Rearchitecture] Check that an install plan exists

tyreer:
# Summary

This work addresses https://github.com/newrelic/newrelic-quickstarts/issues/637

We chose to check if the install plan exists via the repo rather than the catalog service due to the constraints Jerel mentions here: https://github.com/newrelic/newrelic-quickstarts/issues/637#issuecomment-997065715

@caylahamann was the primary engineer on this work, and she'd be a great point person for any questions.

You can see the DevEx PR here: https://github.com/tyreer/newrelic-quickstarts/pull/4

PR #664, opened 2021-12-30
[Repo Rearchitecture] Check that an install plan exists

tyreer:
# Summary
This work addresses #637

We chose to check if the install plan exists via the repo rather than the catalog service due to the constraints Jerel mentions here: #637 (comment)

@caylahamann was the primary engineer on this work, and she'd be a great point person for any questions.

You can see the DevEx PR here: tyreer#4

Note: This replaces an [earlier PR](https://github.com/newrelic/newrelic-quickstarts/pull/660) that had a load of additional commits added onto it, but it's the same change set as that earlier PR, which @zstix [had approved](https://newrelic.slack.com/archives/G01N33Y8P3K/p1640628860005200)


utils/validate-install-plan.js
```diff
@@ -0,0 +1,108 @@
...
+  return configFiles.map(({ filename }) => {
+    const filePath = path.join(process.cwd(), `../${filename}`);
+    const { installPlans } = readQuickstartFile(filePath).contents[0];
+
+    return { filePath, installPlans };
+  });
+};
+
+/**
+ * Gets all the install plans and paths for an array of config files
```
rudouglas, line 44: The description here is the same as `getConfigInstallPlans` so I think that just needs to be altered

utils/validate-install-plan.js
```diff
@@ -0,0 +1,108 @@
...
+      const nonExistentInstallPlans = installPlans.filter(
+        (plan) => !installPlanIds.includes(plan)
+      );
+      return { filePath, installPlans: nonExistentInstallPlans };
+    })
+    .filter(({ installPlans }) => installPlans.length > 0);
+};
+
+/**
+ * Gets all the install plans and paths for an array of config files
```
rudouglas, line 61: Same here, description has been copy/pasted so just needs to be changed

zstix (approved):

utils/validate-install-plan.js
tyreer: Got this over on the main fork PR here: 823dc769

utils/validate-install-plan.js
tyreer: Got this over on the main fork PR here: 823dc76

Issue #635, opened 2021-12-13
[Repo Rearchitecture] Create workflow for validating Install Plans


rudouglas:
## Description

We don't currently have any validation in place for Install Plans added to the Repo. As we are moving to NerdGraph for our single source of truth, we will now need to create/update install plans using a NerdGraph mutation.

The plan is to have an endpoint `nr1CatalogSubmitInstallPlanStep` with a `dryRun` argument for validation, which is currently a WIP

## Acceptance Criteria

- [x] Create Install plan validation workflow
- [x] Create script ~`/utils/validate_install_plans.js` to send install plan data to GraphQL endpoint and receive response and report the success / failure reason back in the workflow
- [x] Handle error messages that are returned
- [x] Create unit tests

### Helpful Links
Schema - https://docs.google.com/spreadsheets/d/1EmAvq4kJUUPlpvJZ-s23uKDk_KCws9VgHzG1bJtwE4Y/edit#gid=0
NerdGraph - https://api.newrelic.com/graphiql?#query=mutation%20%7B%0A%20%20nr1CatalogSubmitMetadata%0A%7D%0A
DevEx Service - https://source.datanerd.us/nr1-dev-experience/catalog-service-elixir

Issue #571, opened 2021-11-17
[Repo Rearchitecture] Update public GitHub PR process to use catalog service mutations for validating data


rudouglas:
## Description

⚠️  This will not replace the current process yet, but will rather work in parallel with the new system so new workflows should be created for this work

The current GitHub PR process uses custom validation scripts to ensure the data is accurate and built correctly. We want to try and leverage the catalog service to perform the validations instead so that as we make tweaks and/or have new data requirements, we make changes in a singular place.

The virtuoso team is working on the mutations right now and will provide a `dryRun` argument for the `submitQuickstart` endpoint. Once https://github.com/newrelic/newrelic-quickstarts/issues/631 is complete we will be able to use this mutation.

### [Validate-packs](https://github.com/newrelic/newrelic-quickstarts/blob/main/.github/workflows/validate_packs.yml)
 - Jobs/Scripts we will need to change to use NerdGraph
      - [Ensure-schema-compliance](https://github.com/newrelic/newrelic-quickstarts/blob/e4ae0a4b5a10fb2da9de324ffdfae8cbbcd206c7/.github/workflows/validate_packs.yml#L6)
          - Script: [/utils/validate_quickstarts.js](https://github.com/newrelic/newrelic-quickstarts/blob/main/utils/validate_quickstarts.js)
      - [Ensure-quickstart-names-ids-are-unique](https://github.com/newrelic/newrelic-quickstarts/blob/e4ae0a4b5a10fb2da9de324ffdfae8cbbcd206c7/.github/workflows/validate_packs.yml#L24) (ID check will not be needed once this is being generated in the NerdGraph Service)
          - Script: [/utils/check_quickstart_uniqueness.js](https://github.com/newrelic/newrelic-quickstarts/blob/main/utils/check_quickstart_uniqueness.js)
 - Waiting for clarification on what other validation this will replace

## Acceptance Criteria
- [ ] Update validation script to be _able_ to run against all quickstarts, but _only_ run against the updated quickstarts at PR time
- [ ] Quickstart validation should now be done via either `create` or `update` endpoints using a `dryRun` argument. https://github.com/newrelic/newrelic-quickstarts/issues/631 will determine which endpoint to send to
[...]

PR #665, opened 2021-12-30
[Repo Rearchitecture]  - Add service validation for quickstarts + install plans

tyreer:
# Summary

This PR addresses both: https://github.com/newrelic/newrelic-quickstarts/issues/571 + https://github.com/newrelic/newrelic-quickstarts/issues/635.

This PR has service dependencies and should remain in draft until the [[NRIO] Migrate quickstart storage to Catalog service database MMF](https://newrelic.atlassian.net/wiki/spaces/~906143720/pages/2368602537/Navigation) is complete.

The code in this PR reflect 2 PRs that have had a round of review from DevEx, so it should be in pretty good shape. That said, while we developed these changes we needed to run the Actions against local NerdGraph on a WIP branch of the catalog service ([ref](https://github.com/tyreer/newrelic-quickstarts/pull/2/files
)). Once the `nr1CatalogSubmitQuickstart` and `nr1CatalogSubmitInstallPlanStep` are completed and deployed, it will be smart to smoke test these workflows and jobs against a few trial PRs to ensure everything is looking good!

## Required GitHub Secret:
- `NR_API_TOKEN`
  - Not sure what the best token to have on the repo will be  

## Notes
- Because this PR is from the `main` branch of our fork, it will also reflect the changes in https://github.com/newrelic/newrelic-quickstarts/pull/664 until that PR is merged in


utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,359 @@
...
+} = require('./helpers');
+const {
+  fetchNRGraphqlResults,
+  translateMutationErrors,
+} = require('./nr-graphql-helpers');
+
+const GITHUB_API_URL = passedProcessArguments[0];
+const GITHUB_REPO_BASE_URL =
+  'https://github.com/newrelic/newrelic-quickstarts/tree/main';
+const GITHUB_RAW_BASE_URL =
```
tyreer, line 23: In the case of a brand new quickstart, URLs build with either the `GITHUB_REPO_BASE_URL` or `GITHUB_RAW_BASE_URL` bases would not exist on the  `main` branch of the repo.

For values on the mutation input object referencing those constants —like the `icon`, `sourceUrl`, or screenshot `url` attributes—would we need to construct URLs based on the PR or a given file's most recent commit?

Something like: https://raw.githubusercontent.com/tyreer/newrelic-quickstarts/80857af8a45a481d144b68bd8206a8d7d804a220/quickstarts/rob-test-qs-invalid/logo.png (from [here](https://github.com/tyreer/newrelic-quickstarts/blob/80857af8a45a481d144b68bd8206a8d7d804a220/quickstarts/rob-test-qs-invalid/logo.png
))

Question maybe is: Is the catalog service reading the contents at that URL and performing validation logic, or simply confirming this is a valid URL?

utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,359 @@
...
+    installPlans,
+    id,
+  } = quickstartConfig.contents[0] || {};
+  const alertConfigPaths = getQuickstartAlertsConfigs(quickstartConfig.path);
+  const dashboardConfigPaths = getQuickstartDashboardConfigs(
+    quickstartConfig.path
+  );
+
+  return {
+    id: id,
```
tyreer, line 125: Pretty sure a brand new quickstart will not have an  `id` in its `config.yml`.  Seeing the addition of a UUID on the `release` workflow, so at the workflow  of candidate PR validation we wouldn't expect an ID: https://github.com/newrelic/newrelic-quickstarts/blob/main/.github/workflows/release.yml#L14. 

That means this `id` value will be `undefined` when sending a validation mutation, so there will always be an error that will make the action / workflow fail on a new quickstart...

It will be good to smoke test this and confirm once the `submitQuickstart` mutation is deployed. 

If the above is correct, we can find a way to handle. Holding off on adding anything preemptively for now.

utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,359 @@
...
+    quickstartMetadata: {
+      alertConditions:
+        alertConfigPaths.length > 0
+          ? adaptQuickstartAlertsInput(alertConfigPaths)
+          : undefined,
+      authors: authors && authors.map((author) => ({ name: author })),
+      categoryTerms: categoryTerms || keywords,
```
tyreer, line 132: We came across a bit of unknown here. Would there ever be a key of `categoryTerms` from an open source `config.yml`?

Looks like the README points contributors exclusively to `keywords` https://github.com/newrelic/newrelic-quickstarts/blob/main/CONTRIBUTING.md#keywords

I think there might just be a bit of overlap in the catalog service ([internal repo ref](https://source.datanerd.us/nr1-dev-experience/catalog-service-elixir/blob/5400c85f2dc8042902e37e27ab0fde4f36280d89/apps/api/lib/api/resolvers/quickstart_metadata.ex#L46-L57)), but if that's not relevant for this repo we might as well remove this

jpvajda:
@tyreer is out this week, but should pick this back up when he's back.

.github/workflows/validate_install_plans.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: 12
+
+      - name: Install dependencies
+        run: cd utils && yarn install
+
+      - name: Validate new files
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
LizBaker, line 23: does this work from a fork? can't remember if this token will be readable

utils/github-api-helpers.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+const fetch = require('node-fetch');
+const parseLinkHeader = require('parse-link-header');
+
+const CONFIG_REGEXP = new RegExp('quickstarts/.+/config.+(yml|yaml)');
```
LizBaker, line 6: i see `QUICKSTART_CONFIG_REGEXP` is referenced on line 59, should this match?
```suggestion
const QUICKSTART_CONFIG_REGEXP = new RegExp('quickstarts/.+/config.+(yml|yaml)');
```

zstix (commented):
Left a few comments but, as far as we can tell, this looks great!

utils/validate_pr_quickstarts.js
jerelmiller: While we don't necessarily need to support that in the config yet, I'd opt for parsing them out of the keywords before submitting to the mutation. The mutation does not associate quickstarts with categories by looking at keywords, so we need to ensure they get submitted as such.

We essentially just need to copy over the `associatedKeywords` from [this list](https://github.com/newrelic/developer-website/blob/develop/src/data/instant-observability-categories.json) and pull out any keyword that matches that list into `categoryTerms` before submitting the mutation.

.github/workflows/validate_install_plans.yml
zstix: Looks like we can enable this: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

.github/workflows/validate_install_plans.yml
```diff
@@ -0,0 +1,28 @@
+name: Validate install plans
+
+on: [workflow_dispatch, pull_request]
```
zstix, line 3: I think this is what we need to do to ensure that this runs with the secrets / version of the workflow that we have in the base branch.

```suggestion
on: [workflow_dispatch, pull_request_target]
```

utils/validate_pr_quickstarts.js
tyreer: Thanks @jerelmiller. 

Could I confirm I'm following the suggestion right before implementing this:

1. User submits a config with `keywords`
2. Workflow looks at those keywords from the user's config and match them to the category `value` via the `associatedKeywords` set
3. Workflow sends only the set of any category `value` strings in the `categoryTerms` key of the mutation variables this function is building up

Does that sound right?

utils/validate_pr_quickstarts.js
tyreer: Not seeing any objections, so I'm going to stick with the `/main/` path

utils/validate_pr_quickstarts.js
jerelmiller: Thats correct!

utils/validate_pr_quickstarts.js
nr-kkenney: did some quick testing in staging. we can't supply empty, or an id with an invalid "shape" -- looks like it has to be guid / uuid format. however, as long as we do that, the id can be any value and we can validate with it.

so the process would be:
* grab id from quickstart
* if its new and id doesn't exist, replace it with a default id. could be `00000000-0000-0000-0000-000000000000` as an example
* validate using that as the id in the query

utils/github-api-helpers.js
tyreer: Yes, thanks! c5c2c71c

.github/workflows/validate_install_plans.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Install dependencies
+        run: cd utils && yarn install
+
+      - name: Validate new files
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NR_API_URL: 'https://api.newrelic.com/graphql'
+          NR_API_TOKEN: ${{ secrets.NR_API_TOKEN }}
+        run: 
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
```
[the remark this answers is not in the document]
nr-kkenney: looks like this is one the default environment variables GH sets for us. From the docs:
> 
GITHUB_REPOSITORY | The owner and repository name. For example, octocat/Hello-World.
so I think we're set to use it 👍

utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,359 @@
...
+
+/**
+ * Looks up corresponding quickstart config files for quickstarts known to have changes in a PR.
+ * @param {Set} quickstartDirectories - Set of unique quickstart directories.
+ * @return {Array} Collection of config file paths.
+ */
+const getQuickstartConfigPaths = (quickstartDirectories) => {
+  const allQuickstartConfigPaths = findMainQuickstartConfigFiles();
+
+  return [...quickstartDirectories].reduce((acc, quickstartDirectory) => {
```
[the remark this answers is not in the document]
tyreer: `quickstartDirectories` is a `set` here, so we need to convert to an `array`. 

Might change to make it more explicit:

```
return Array.from(quickstartDirectories).reduce((acc, quickstartDirectory)
```

utils/validate_pr_quickstarts.js
tyreer: Got it, I've implemented a fallback UUID here for validation of new quickstarts: c2703ffc

Thinking out loud through an edge case / tradeoff: With this change, any modification to an existing quickstart config where someone has accidentally removed the assigned ID in the `config.yml` would not benefit from validation of that missing field. Suppose in that case a brand new quickstart would be made if the PR is merged, which seems reasonable. I don't see how this PR workflow could know if a given quickstart without an `id` in its `config.yml` is a new addition as opposed to a modified existing config.

Feels okay to move forward with the mock UUID for validation purposes.

.github/workflows/validate_install_plans.yml
nr-kkenney: im not 100% sure of why things work atm 🙈 

but i dont think we need this. as an example, this workflow (or the previous iteration) didn't have pull_request_target, but it has been running like that all this time. i just tested it out with a little change on a fork i made, and it works there.

i think this running & working comes from our actions approval setting underneath `Fork pull request workflows from outside collaborators`. So my theory is because we have a setting checked there, we can use pull_request instead of pull_request_target for collaborators we approve.

.github/workflows/validate_install_plans.yml
nr-kkenney: i think this would work as is. i think the token would have read permissions by default, which the script would be able to use since its only querying for data.

utils/validate_pr_quickstarts.js
nr-kkenney: talked with @zstix and it sounds like we're cool with this for now. i created [a ticket](https://github.com/newrelic/newrelic-quickstarts/issues/698) for this edge case so we can keep track of it.

utils/validate_pr_quickstarts.js
tyreer: Cool, this is where I landed: 0a9e41ab6724b5cb88f5a09b185b424f30bf50aa

Ended up removing `categoryTerms` as a value taken from the user-supplied config since we're not expecting them to supply those categories and building the array of categories based on keywords.

utils/validate_pr_quickstarts.js
tyreer: Great, thank you!

.github/workflows/validate_install_plans.yml
tyreer: I've got a [testing PR opened on our fork](https://github.com/tyreer/newrelic-quickstarts/pull/2) that I've been using to run the script in a GH context. So long as we provide a `NR_API_TOKEN` in the fork secrets, it picks up `GITHUB_TOKEN` and can read from the GH API as needed. 

I assume that's actually our fork's GH token, but regardless I think this should be all good. As @nr-kkenney mentioned the key is just being used to read what files are in the current PR.

Here's a job run on the fork test PR where the validation is failing as expected, but the script itself is working: https://github.com/tyreer/newrelic-quickstarts/runs/4855812719?check_suite_focus=true

utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,359 @@
...
+      }
+  }
+`;
+
+/**
+ * Gets the quickstart portion of a given file path.
+ * @param {String} filePath - Full file path of a file in a quickstart.
+ * @param {String} targetChild - Node in file path that should be preceded by a quickstart directory.
+ * @return {String} Node in file path of the quickstart.
+ */
```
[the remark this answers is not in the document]
tyreer: I can see how an example would be helpful. Added here: 4557da8d

nr-kkenney (approved):

utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,359 @@
...
+ * Looks up corresponding quickstart config files for quickstarts known to have changes in a PR.
+ * @param {Set} quickstartDirectories - Set of unique quickstart directories.
+ * @return {Array} Collection of config file paths.
+ */
+const getQuickstartConfigPaths = (quickstartDirectories) => {
+  const allQuickstartConfigPaths = findMainQuickstartConfigFiles();
+
+  return [...quickstartDirectories].reduce((acc, quickstartDirectory) => {
+    const match = allQuickstartConfigPaths.find((path) =>
+      path.split('/').includes(quickstartDirectory)
```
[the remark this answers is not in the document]
tyreer: This is a great catch! The logic would not handle cases like that at all 😅  

I've updated here: ce7c68c1

nr-kkenney (dismissed):

Issue #605, opened 2021-11-30
[Repo Rearchitecture] `Ensure commits are conventional` workflow is silently failing


aswanson-nr:
## Description

The `Ensure commits are conventional` workflow is reporting `success` on pull requests that it should be reporting `failure`

```bash
2021-11-29T18:48:48.9479774Z ##[group]Run ./utils/conventional-commits.sh
2021-11-29T18:48:48.9480642Z [36;1m./utils/conventional-commits.sh[0m
2021-11-29T18:48:48.9532987Z shell: /usr/bin/bash -e {0}
2021-11-29T18:48:48.9533443Z ##[endgroup]
2021-11-29T18:48:48.9648177Z fatal: ambiguous argument 'origin..HEAD': unknown revision or path not in the working tree.
2021-11-29T18:48:48.9649722Z Use '--' to separate paths from revisions, like this:
2021-11-29T18:48:48.9651207Z 'git <command> [<revision>...] -- [<file>...]'
2021-11-29T18:48:48.9756282Z Post job cleanup.
```
## Steps to Reproduce
* See https://github.com/newrelic/newrelic-quickstarts/runs/4357923215?check_suite_focus=true
* Associated PR https://github.com/newrelic/newrelic-quickstarts/pull/602

## Expected Behavior
* If the commits do not follow the conventional commit syntax, then this check should fail.

## Acceptance Criteria
* [ ] Remove conventional commit code/workflow
* [ ] Update the release workflow to work without conventional commits. When a PR is merged, always publish a release.
* When the OILS is deprecated, remove release code all-together (for future ticket)

zstix:
We have [a script](https://github.com/newrelic/newrelic-quickstarts/blob/main/utils/conventional-commits.sh) that is supposed to check for this. It might be harder to solve this now that there are commits in the repo that don't follow the conventional commit format.

zstix:
Conventional commits are needed for releases, which will not be necessary when we move from a _pull_ to a _push_ model (PRs into main will send a mutation straight to NerdGraph). We wont need releases so we can remove this.

jpvajda:
I confirmed with @jerelmiller, this isn't blocked. 

_We should be ok to finish all work on DevEn side regardless of the status of the service, as long as we don’t yet switch to using the service as the primary means of deploying/validating_

jpvajda:
We discussed that this would just be a removal of the conventional commit code/workflow but @aswanson-nr was going to add some clarifications.

[...]

PR #681, opened 2022-01-11
chore: remove conventional commit checks

rudouglas:


zstix:
Closes https://github.com/newrelic/newrelic-quickstarts/issues/605

rudouglas:
Going to test this with a fork

aswanson-nr (changes requested):
Changes look good, however it still appears to be analyzing commits for the conventional commit syntax to decide whether it should generate a release. I forked the repository to test it, here is a [release workflow run](https://github.com/aswanson-nr/newrelic-quickstarts/runs/4796968464?check_suite_focus=true) from that fork.

PR #695, opened 2022-01-14
feat: add sourceUrl to dashboard and alert assets

rudouglas:


aswanson-nr (approved):
We could add some unit tests for this function, but I don't think that's necessary. Looks good to me!

Issue #670, opened 2022-01-03
[Repo Rearchitecture] Review and merge DevEx fork for quickstart validations


aswanson-nr:
## Summary
After DevEx finishes the work for validating quickstarts against nerdgraph, we will need to sit down and review the changes with them before merging.

* [Fork PR](https://github.com/newrelic/newrelic-quickstarts/pull/665/files)
## Acceptance Criteria 

- [x] Work is handed off to the DevEn team
- [ ] Work is merged into `main`

nr-kkenney:
didn't realize we also had this ticket 🙈 

as an update, the PR is mostly good. the one thing we didn't realize was that the current workflow doesn't work correctly for contributions from forks. secrets are not available, so we can't call the api (as part of `pull_request` trigger) when the contribution comes from a fork.

this means we have to evaluate a different way to move forward. additionally, it seems that no matter what route we choose, we can't get everything we want currently 🙈 

something has to be sacrificed in order for us to move forward

nr-kkenney:
we were on board to try using `workflow_run` and having a workflow trigger another workflow. i've implemented that change here: https://github.com/newrelic/newrelic-quickstarts/pull/665/commits/b50e12608e67f8f2d3714a5bf7445a7c562f455b.

PR #705, opened 2022-01-20
merge validation feature branch into main

nr-kkenney:
## Summary

This includes all the validation updates which were introduced by `dev-ex`. Those changes have already been reviewed by their merge into the feature branch.

The primary change set to lookout for and review are the workflows -- anything underneath `.github`, which I added somewhat independently from any review 🙈 

## Links
Relates to: #670


.github/workflows/validation_gate.yml
```diff
@@ -0,0 +1,25 @@
+# This looks like a confusing workflow, but it has a purpose (✿ ◕‿◕)
+# It functions kind of as a workflow shim. We can't setup other important workflows to run on pull_request (due to forks), so we setup this workflow to be triggered by pull_request and it is always able to run.
+# When it is finished running, it triggers the other important workflows -- validate_packs, validate_install_plans.
+# Having it setup this way allows us to have manual approval for this workflow run, and then the subsequent triggered runs can access secrets in the repository for validation (even for forked PRs).
+name: Validation Gate
+
+on: 
+  pull_request:
+    branches:
+      - 'main'
```
aswanson-nr, line 10: Are we sure we want this to only run on pull requests into `main`? I could see us wanting validation when merging into branches, such as feature branches

.gitignore
```diff
@@ -2,3 +2,5 @@ utils/node_modules
 
 # IDE files
 .DS_Store
+settings.json
+.env
```
aswanson-nr, line 6: Adding a star pattern here will also take care of files like `.env.development` 
```suggestion
.env*
```

utils/instant-observability-categories.json
```diff
@@ -0,0 +1,72 @@
+[
```
aswanson-nr, line 1: I'm not sure why this file is here (I may not have gotten to that part of the PR yet), quickstarts are associated with their categories via the keywords they use, so I can't see any obvious use for this. Especially since when we move over to SSR, we will be getting all of these from the API, so I could see this getting out of date very quickly.

utils/validate_pr_quickstarts.js
```diff
@@ -0,0 +1,417 @@
...
+
+  return {
+    id: id ? id : MOCK_UUID,
+    quickstartMetadata: {
+      alertConditions:
+        alertConfigPaths.length > 0
+          ? adaptQuickstartAlertsInput(alertConfigPaths)
+          : undefined,
+      authors: authors && authors.map((author) => ({ name: author })),
+      categoryTerms: getCategoriesFromKeywords(keywords),
```
aswanson-nr, line 152: @jerelmiller or @tyreer, does nerdgraph expect us to provide the categories? I vaguely remember talking about this. Initially keywords were associated with particular keywords so that we could update the quickstarts and categories independently of each other. If we are specifying categories explicitely then we could go in and completely remove keywords

aswanson-nr (commented):
Looks almost ready to go! I had a few comments and some of them could just come from misunderstandings on my part!

This PR deletes the `quickstarts/react-native/dashboards/ReactNative.png`, let's be sure we want to remove that since it seems unrelated to the rest of these challenges

utils/validate_pr_quickstarts.js
tyreer: Related but not an answer per se, just a bit of context buried in a comment thread on this other PR: https://github.com/newrelic/newrelic-quickstarts/pull/665#discussion_r776896725

tyreer: Rereading that thread though, I believe NerdGraph does expect categories. But we're anticipating users would just supply keywords via config, which we map to the categories with this function.

.github/workflows/validate_packs.yml
```diff
@@ -1,110 +1,307 @@
...
+        id: validation
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NR_API_URL: "https://api.newrelic.com/graphql"
+          NR_API_TOKEN: ${{ secrets.NR_API_TOKEN }}
+        run: |
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
+          cd utils && yarn validate-pr-quickstarts $URL
+
+      - name: Add commit status
```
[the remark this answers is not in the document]
nr-kkenney: i dont think there is a way to leverage that for individual steps. we may be able to DRY up all the jobs, but that is not a change im convinced in making, at least not as part of this PR unless we are certain we want & need it.

i would rather that be in a separate ticket devote some research & planning into.

if im imagining it right we would have some generic javascript script runner workflow, and then each job simply calls that workflow with the script to run and the scripts arguments.

utils/validate_pr_quickstarts.js
aswanson-nr: Got it! That makes total sense!

jerelmiller: > does nerdgraph expect us to provide the categories?

**Short answer**: yes

**Long answer**: Since we are designing the API to be accessible to more than just this repo, we made the decision to enforce the requirement that in order to associate a quickstart with a category, it would need to be sent in the mutation as part of the `categoryTerms` field. This makes it obvious to a consumer of the mutation which are actual intended keywords (free text) vs the predefined set of categories (restricted set).

>  Initially keywords were associated with particular keywords so that we could update the quickstarts and categories independently of each other

This is still true, though we call these "category terms". The list of category terms is made up of those list of predefined keywords that we designated as categories. You can see this in NerdGraph using [this query](https://api.newrelic.com/graphiql?#query=%7B%0A%20%20actor%20%7B%0A%20%20%20%20nr1Catalog%20%7B%0A%20%20%20%20%20%20categories%20%7B%0A%20%20%20%20%20%20%20%20displayName%0A%20%20%20%20%20%20%20%20terms%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A). 


> If we are specifying categories explicitely then we could go in and completely remove keywords

I wouldn't completely remove keywords. Keywords are still useful as a free form text field. These keywords are matched when you type something into the search bar when trying to filter down the quickstarts. Keywords are still useful to provide some additional metadata to give you better search results. They just are free-from rather than a predefined set of values.

We don't necessarily need to go update all the quickstart definitions yet, and this is where that `utils/instant-observability-categories.json` file comes into play. Rather than needing to go update the quickstarts right now (introducing a new field to the schema, mass updates on the config files, etc), the idea was that the script could go looks at the configured keywords and pluck the keywords that match the predefined `categoryTerms`. The script would then submit the quickstart with those `categoryTerms` in the mutation. We just need to know which keywords are considered "category terms" in order to compare the keywords and pluck the right values. Make sense?

This buys us time to allow authors to continue to use `keywords`, but maintain the ability to separate the concerns on the API side. I don't see this as a long-term thing, just a stop-gap until we can get a `categoryTerms` field introduced to the yaml configuration. Once we can get a proper field in the config, we can remove this special treatment.

Does that help?

.github/workflows/validation_gate.yml
nr-kkenney: sound good to me. ill make this change.

utils/instant-observability-categories.json
tyreer: I think I brought this over initially to get the categories from user-provided keywords.

If there's another source of the mapping from keywords to categories that's centralized, that would be great to use. I had the same "falling out of sync" concern when I copied this JSON from here: https://github.com/newrelic/developer-website/blob/develop/src/data/instant-observability-categories.json

.github/workflows/validate_packs.yml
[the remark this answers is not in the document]
aswanson-nr: Sounds good!

utils/nr-graphql-helpers.js
```diff
@@ -0,0 +1,104 @@
...
+ * @param {String[] | undefined} configKeywords  - An array of keywords specified in a quickstart config.yml
+ * @returns {String[] | undefined } An array of quickstart categories
+ */
+const getCategoriesFromKeywords = (configKeywords = []) => {
+  const categoriesFromKeywords = instantObservabilityCategories.reduce(
+    (acc, { associatedKeywords, value }) => {
+      if (
+        associatedKeywords.some((keyword) => configKeywords.includes(keyword))
+      ) {
+        acc.push(value);
```
aswanson-nr, line 90:
```suggestion
        acc.push(associatedKeywords);
```

jerelmiller: We should actually only push on the subset of `keywords` on the quickstart that also appear as part of the category. We don't necessarily want to push on the whole set of associated keywords from the category.

For example, the `Infrastructure & OS` category has both `os` and `infrastructure` `categoryTerms`. A quickstart might only include the `infrastructure` keyword, but that doesn't also mean its `os`. In this case, we'd want to detect that `infrastructure` is one of the keywords associated with a category, and pass only that keyword via `categoryTerms`.

Does that make sense?

aswanson-nr: I'm still not convinced categoryTerms provides any value, but those changes make sense!

jerelmiller: Thats the only away to associate the quickstart with a category 😄 

The idea is that `keywords` is free-form text to aide is searchability. These can be anything the user wants them to be. `categoryTerms` are meant to relate to categories, which is a predefined set of values we are using as a filtering mechanism. They are similar, but vastly different scopes.

Since categories and keywords are 2 distinct things, we wanted that reflected in the set of fields. By conflating keywords and categories into a single field called `keywords`, its difficult to know which are the "free form" values and which are supposed to represent a relationship to a category. 

Does that help?

jerelmiller: Also to add on... why `categoryTerms` and not the categories themselves?

We wanted to have the power to compose smaller units (category terms) into a broader bucket (category). Let's say tomorrow we wanted to break apart `Infrastructure & OS` and create separate `Infrastructure` and `OS` categories. Or lets say we wanted to combine `Kubernetes & containers` with `OS`. What would that mean for existing quickstarts that are attached to these categories?

By utilizing category terms as the composable unit, this allows us to perform the above changes without needing to go update a bunch of quickstarts. Quickstarts attached to the `os` category term, would automatically get bucketed with the updated categories. No need perform a bunch of updates anytime we need to make changes to the categories.

Hopefully this helps as well understand why we chose that direction!

tyreer: Pretty sure this was my code from the [earlier fork PR](https://github.com/newrelic/newrelic-quickstarts/pull/665#discussion_r776896725). 

My understanding is that DevEn are working through these changes. Just being explicit that I'm not planning to pick up the suggested change here and trust the DevEn rock stars are working through revisions. Thanks for refining this work! 🎸 

_(but also, do let us know if you'd like DevEx to hop on here and make the changes, happy to help!)_

nr-kkenney: changes made. @jerelmiller @aswanson-nr @tyreer if yall feel like 👀

utils/instant-observability-categories.json
aswanson-nr: Keeping this for backwards compatibility!

aswanson-nr (approved):
:shipit:

utils/nr-graphql-helpers.js
aswanson-nr: Changes look good to me!

Issue #636, opened 2021-12-13
[Repo Rearchitecture] Use new service to create/update Install Plans


rudouglas:
### Description

We will need to use the new service to create/update install plans based on contributions. This will require a new Job and script. We need to submit the install plan to `nr1CatalogSubmitInstallPlanStep`, we don't need to worry about separate jobs for create/update as this will be handled under the hood in the service.

### Acceptance Criteria 

> deletion of install plans is not in scope as deleting anything from the quickstart catalog isn't currently supported. 

- [ ] DevEx will have a [fork](https://github.com/newrelic/newrelic-quickstarts/pull/665 this functionality we want to confirm is available prior to starting this work.
- [ ] Create a new Job in the [release.yml](https://github.com/newrelic/newrelic-quickstarts/blob/e4ae0a4b5a10fb2da9de324ffdfae8cbbcd206c7/.github/workflows/release.yml) that runs a new script to submit the install plan to `nr1CatalogSubmitInstallPlanStep`
 - [ ] Should send the configuration details to the new NerdGraph mutation endpoint
 - [ ] Should handle errors returned from the service
 - [ ] Should include unit tests

We will also need to take into account our 3 environments which will need to be released at the same time:
 - EU Production
 - Production
 - Staging

#### Helpful Links

Schema - https://docs.google.com/spreadsheets/d/1EmAvq4kJUUPlpvJZ-s23uKDk_KCws9VgHzG1bJtwE4Y/edit#gid=0
NerdGraph - https://api.newrelic.com/graphiql?#query=mutation%20%7B%0A%20%20nr1CatalogSubmitMetadata%0A%7D%0A
DevEx Service - https://source.datanerd.us/nr1-dev-experience/catalog-service-elixir

jpvajda:
From this [comment](https://github.com/newrelic/newrelic-quickstarts/issues/631#issuecomment-992745503) it sounds like there are few changes to the approach @rudouglas

jerelmiller:
To make this concrete:

This will use a single mutation. No need to distiguish between create/update. Since the client is generating the `id`, you can use the `nr1CatalogSubmitInstallPlanStep` mutation. We'll handle the create or update under the hood.

jpvajda:
I confirmed with @jerelmiller, this isn't blocked. 

_We should be ok to finish all work on DevEn side regardless of the status of the service, as long as we don’t yet switch to using the service as the primary means of deploying/validating_

Issue #567, opened 2021-11-16
[Repo Rearchitecture] Use DevEx service to create new and update existing quickstarts


rudouglas:
### Description

⚠️  This will not replace the current process yet, but will rather work in parallel with the new system so new workflows should be created for this work

Quickstarts will be created/updated via NerdGraph.

We will use the `submitQuickstart` endpoint to update the data for that quickstart. The endpoint will return the quickstart details but it may also return errors so we need to be able to handle them effectively. We will need a list of potential errors returned by the endpoint and implement things like retry logic etc.

We will also need to take into account our 3 environments which will need to be released at the same time:
 - EU production
 - Production
 - Staging

NOTE: This will not replace the current process yet, but will rather work in parallel with the new system so new workflows should be created for this work

### [Job-generate-release](https://github.com/newrelic/newrelic-quickstarts/blob/832f28ca88b23e5bd678019340b02c58d34037f9/.github/workflows/release.yml#L190)

 - New Job/Script needed
    - `update-quickstart` - This should be in [release.yml](https://github.com/newrelic/newrelic-quickstarts/blob/e4ae0a4b5a10fb2da9de324ffdfae8cbbcd206c7/.github/workflows/release.yml) and run the new script

### Acceptance Criteria

- [x] DevEx will have a [fork](https://github.com/newrelic/newrelic-quickstarts/pull/665) this functionality we want to confirm is available prior to starting this work.
 - [x] Create a new Job in the [release.yml](https://github.com/newrelic/newrelic-quickstarts/blob/e4ae0a4b5a10fb2da9de324ffdfae8cbbcd206c7/.github/workflows/release.yml) that runs a new script ~`update-quickstart`
 - [x] The script should identify an updated quickstart when the `config.yml` has an existing `id` field.
 - [x] Should send the configuration details to the DevEx `submitQuickstart` NerdGraph endpoint
 - [x] Should handle errors returned from the service
 - [x] Should include unit tests

#### Helpful Links

Schema - https://docs.google.com/spreadsheets/d/1EmAvq4kJUUPlpvJZ-s23uKDk_KCws9VgHzG1bJtwE4Y/edit#gid=0
NerdGraph - https://api.newrelic.com/graphiql?#query=mutation%20%7B%0A%20%20nr1CatalogSubmitMetadata%0A%7D%0A
DevEx Service - https://source.datanerd.us/nr1-dev-experience/catalog-service-elixir

rudouglas:
[...]

PR #711, opened 2022-01-21
Submit install plans and quickstarts

LizBaker:
Adds a new workflow upon PR merges to main for submitting install plans and quickstarts that have been updated or created

### 🚨 We'll want to check in with DevEx before merging
** update: devex is ready for us to write to prod/eu

We're using env vars for the api urls and tokens. since prod isn't ready we're thinking of setting all of them to staging. this means quickstarts/install plans would get submitted to staging three times, but also means we wouldnt need to PR to turn on prod and eu when the time comes 🤷‍♀️ 

closes https://github.com/newrelic/newrelic-quickstarts/issues/567
closes https://github.com/newrelic/newrelic-quickstarts/issues/636


.github/workflows/pr-merged.yml
```diff
@@ -0,0 +1,192 @@
...
+
+      - name: Install dependencies
+        run: cd utils && yarn install
+      - name: Update install plans - staging
+        env:
+          NR_API_URL: ${{ secrets.NR_API_URL_STAGING }}
+          NR_API_TOKEN: ${{ secrets.NR_API_TOKEN_STAGING }}
+        run: |
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
+          DRY_RUN=false
```
zstix, line 130: If we wanted to reduce some of the duplication, we might be able to set these variables at the workflow level (rather than at the step level).

zstix (approved):
A couple of super minor formatting things, but this looks great!

.github/workflows/release.yml
```diff
@@ -10,94 +10,6 @@ env:
   THIRD_PARTY_GIT_AUTHOR_NAME: nr-opensource-bot
 
 jobs:
-  generate-uuid:
```
aswanson-nr, line 13: We will need to keep this workflow for now, the service does not generate the UUID on the server side. This is mainly due to it running on multiple environments and the UUID needing to be consistent across them

.github/workflows/pr-merged.yml
LizBaker: so true!

.github/workflows/release.yml
LizBaker: this just moves the job to the new pr-merged workflow, it will still generate and commit UUIDs

.github/workflows/pr-merged.yml
```diff
@@ -0,0 +1,188 @@
+name: Update quickstarts
+
+env:
+  THIRD_PARTY_GIT_AUTHOR_EMAIL: opensource+bot@newrelic.com
+  THIRD_PARTY_GIT_AUTHOR_NAME: nr-opensource-bot
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aswanson-nr, line 6: We also need the opensource bot token, let's call that out here too!

aswanson-nr (approved):
Looks good to me! Let's get this tested on a fork before we merge

aswanson-nr (approved):
:shipit:

PR #747, opened 2022-01-27
feat: new quickstart Slack for CodeStream

jcountsNR:
# Summary

Documentation only quickstart for Slack CodeStream. NO Dashboard, Alerts, or data integrations.

## Pre merge checklist

- [ X] Did you check your descriptive content for voice, tone, spelling and grammar errors?


jpvajda (changes requested):
Is there a reason why this quickstart wouldn't be stored in the /codestream directory as it relates specifically to that product area? I noticed you created a new directory for it.

jcountsNR:
Yes it probably should, there are about 6 more I need to after this as well related to CodeStream, so it makes sense to start a directory. Also, I'm having a hard time understanding why it is failing, there is no install id mentioned.

jpvajda:
@jcountsNR I'm looking into it now, it may be a bug 😅

nr-kkenney:
@jcountsNR the failure for validation / install plan ids exist looks like a bug on our end 🙈 

we did fix the `Validation / Quickstart schema compliance` check. so if you rebase / merge main, you should see that succeed on the next run.

ill take a look at the install plan id check failure.

nr-kkenney:
@jcountsNR fix should be in for the install plan validation. so if you merge main again, that check should succeed.

i see the `Validation / Quickstart schema compliance` check failed again, but this time its checking against your submission and not the error we were seeing before.

jcountsNR:
@nr-kkenney looks like it's passing now. Thanks!

jpvajda (approved):

PR #751, opened 2022-01-28
Feat/add cloud install plans

natecanfield822:
# Summary

This is adding the nerdletState install plan directive field, applying it to one existing AWS install plan, and creating new install plans for AWS, GCP, and Azure leveraging it.

These install plans have been applied to the majority of AWS, GCP, and Azure quickstarts as well as updating their existing documentation links to point to documentation rather than the nerdlet opened by this install plan.


aswanson-nr:
@natecanfield822 Just FYI we removed the files you edited in `/docs` in favor of an auto generated doc from the graphql schema. I removed that file when I merged in main, so this should be good to go!

PR #752, opened 2022-01-28
Maint/add operator baseline

natecanfield822:
# Summary

Adding the "ABOVE" operator explicitly to existing baseline alerts and its template


aswanson-nr:
Hey @natecanfield822 We just merged a fix for the `featured` error you're seeing, if you get the latest from `main` you should get accurate validations!

aswanson-nr (approved):

PR #756, opened 2022-01-28
fix for validating quickstarts with no install plans

nr-kkenney:
# Summary

This PR should resolve the error we were seeing for PR's submitting quickstarts with no install plans. Previously, install plans being missing from the config would cause an error, but is actually a valid quickstart submission.

Also:
* added a unit test for this case.
* updated install plan validation job to use node 16, which lets us use optional chaining & other snazzy additions.

# Links
Resolves: #756


josephgregoryii (approved):
LGTM code wise and seeing all the checks pass

PR #763, opened 2022-01-28
fix error logging

nr-kkenney:
# Summary

We were only printing the remaining errors (after we removed some) in certain cases. Now, even when we filter out errors for validation, we should still see all the errors we encountered.


nr-kkenney:
didnt branch this correctly.

PR #764, opened 2022-01-28
fix for error logging

nr-kkenney:
# Summary

We currently filter out install plan errors we encounter, but still wish to have those logged out. In this instance however, we weren't logging them out 🙈 

With this change, we now log all errors out, including errors relating to install plans not existing.

Output looks like:
```
ERROR: The following errors occurred while validating: quickstarts/elasticsearch/config.yml
- Argument "quickstartMetadata" has invalid value $quickstartMetadata.
In field "summary": Expected type "String!", found null.

ERROR: The following errors occurred while validating: quickstarts/fastly/config.yml
- quickstartMetadata/dashboards/rawConfiguration/pages/widgets/linkedEntityGuids: `linkedEntityGuids` must not be set
- quickstartMetadata/alertConditions/rawConfiguration/terms/operator: `operator` can't be blank
DEBUG: The following are install plan errors that occured while validating: quickstarts/fastly/config.yml and can be safely ignored.
- quickstartMetadata/installPlanStepIds: `installPlanStepIds` contains an install plan step that does not exist: 'third-party-fastly'
```

### DIFF
--- a/utils/nr-graphql-helpers.js
+++ b/utils/nr-graphql-helpers.js
@@ -59,11 +59,12 @@ const fetchNRGraphqlResults = async (queryBody) => {
  * Handle errors from GraphQL request
  * @param {Object[]} errors  - An array of any errors found
  * @param {String} filePath  - The path related to the validation error
+ * @param {Object[]} installPlanErrors - Array of install plan errors which are handled differently
  * @returns {void}
  */
-const translateMutationErrors = (errors, filePath) => {
+const translateMutationErrors = (errors, filePath, installPlanErrors = []) => {
   console.error(
-    `ERROR: The following errors occurred while validating: ${filePath}`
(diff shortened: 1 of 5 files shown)


clarkmcadoo (dismissed):

utils/nr-graphql-helpers.js
```diff
@@ -59,11 +59,12 @@ const fetchNRGraphqlResults = async (queryBody) => {
  * Handle errors from GraphQL request
  * @param {Object[]} errors  - An array of any errors found
  * @param {String} filePath  - The path related to the validation error
+ * @param {Object[]} installPlanErrors - Array of install plan errors which are handled differently
```
aswanson-nr, line 62:
```suggestion
 * @param {Object[]}  [installPlanErrors=[]] - Array of install plan errors which are handled differently
```

aswanson-nr (approved):

PR #765, opened 2022-01-31
feat: generate schema docs from GraphQL

rudouglas:
<details><summary>1st attempt</summary>This queries GraphQL for Endpoint information in order to be able to create Docs from them.

Currently, it looks like there's no easy way to introspectively query all associated endpoints at once. Due to the fact these endpoint names may change in the future, or new ones be introduced, it would be better to query the Base endpoint first, and then recursively query all endpoints associated with it. 

The base endpoint is `Nr1CatalogQuickstart`. Performing an introspection query on this does not give you details on all connected endpoints. As a result we have to search through this data, which could be deeply nested, for endpoint names that start with `Nr1Catalog`. There are `~35` individual endpoints that we need data on, these are listed in the `example-output.json` file in the PR, along with examples of what the data looks like returned from GraphQL.

There doesn't seem to be any packages we can use to turn these into Markdown files, all the packages I could find only work with `__schema` introspection queries but, that only queries the root Query</details>

I have figured out a single query that we can use to get all of the data we need. The initial query sends us every single queryable endpoint in New Relic GraphQl so we have to filter that down to just `Nr1Catalog*` endpoints. But we can now use `graphql-markdown` to auto-generate a Markdown doc from these results :D


utils/graphql-schemas/generate-schema-docs.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {Object} data - The data to filter
+ * @returns {Object} The filtered data in schema form for graphql-markdown
+ */
+const filterData = (data) => {
+  const {
+    data: {
+      __schema: { types },
+    },
+  } = data;
+  const endpoints = types.filter((type) => type.name.startsWith('Nr1Catalog'));
```
aswanson-nr, line 25: This seems to include a lot of the extra stuff under the `Nr1Catalog` namespace, maybe we could filter it down a bit more so it's just install plans and quickstarts?
Something like this might work to filter this down to just what we want
```suggestion
  const endpoints = types.filter((type) => type.name.startsWith('Nr1CatalogQuickstart') || type.name.startsWith('Nr1CatalogInstall'));
```

rudouglas: I guess... ` && type.name !== 'KitchenSink'` 😂

aswanson-nr (dismissed):
Looks good to me! I think we may want to move to a manually defined schema doc in the future, since the nerdgraph naming conventions are a bit different than the yaml definitions. However, this totally works to replace the current auto-generated docs.
I'd just like to get @jpvajda's sign off on the generated doc before we merge

utils/graphql-schemas/schema-docs.md
```diff
@@ -0,0 +1,2306 @@
...
+</tr>
+</tbody>
+</table>
+
+### Nr1CatalogQuickstart
+
+Information about the quickstart
+
+---
+**NR Internal** | [#help-nr1-dev-experience](https://newrelic.slack.com/archives/CPE597DNY) | visibility(customer)
```
jerelmiller, line 287: We may not want to expose our slack channels and field visibility publicly

jerelmiller (dismissed):
Can we strip out internal info from the resulting markdown file?

utils/graphql-schemas/schema-docs.md
jerelmiller: One idea to make this easy is to create a "regular" account which is non NR admin, create a key from there, and use that key to generate these docs. NerdGraph will strip this out for you in that case so that you don't need to worry about creating regexes to strip this out yourself.

nr-kkenney:
quick question, do we have both sets of documentation in the repo? 

should we replace the content in ./docs with the new documentation?

jpvajda:
@nr-kkenney  

> should we replace the content in ./docs with the new documentation?

Maybe? I guess I'd be curious in what's the difference will be in documentation, will it be similar (better || worse)  in what we currently generate in /docs with the script you created?

nr-kkenney:
john and i thought this new documentation was better than the one we currently have, so we're in favor of replacing the existing docs.

did we want to delete the docs we currently have as part of this pr, as well as update any readme and other tasks, or a separate PR?

zstix:
TODO: move into the `docs/` directory.

nr-kkenney (dismissed):
lgtm

.github/workflows/release.yml
```diff
@@ -110,17 +110,17 @@ jobs:
           ref: 'main'
           fetch-depth: 0
 
-      - name: Setup Python
-        uses: actions/setup-python@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
         with:
-          python-version: '3.8'
+          node-version: 12
```
nr-kkenney, line 116: we should use latest / 16 unless we have a reason for 12.

rudouglas: yes please

nr-kkenney (approved):

PR #779, opened 2022-02-02
chore: Config field updates to icon and description

rudouglas:
### DO NOT MERGE
This updates a couple of fields in the quickstart configs:

 - `logo` changed to `icon` in the main `config.yml`
 - `details` changed to `description` in alerts config files
     - Duplicated `description` field removed

This also required changes to our __tests__, scripts. These run successfully locally, but the checks are failing in the PR because they run on the `main` branch. 

Also includes updates to the Docs.

To review more easily, you can use the `Filter By File` functionality in GitHub to remove the `yml` files as most of these are small changes to quickstart configs


jerelmiller:
Will this affect the OILS service? Since the catalog service is still pulling from OILS until the cutover, let's make sure either the OILS service is aware of this change, or hold off on merging this until the catalog service is reading from its own db

aswanson-nr:
> Will this affect the OILS service? Since the catalog service is still pulling from OILS until the cutover, let's make sure either the OILS service is aware of this change, or hold off on merging this until the catalog service is reading from its own db

Good call @jerelmiller, we will hold off on this until we are okay with removing the OILS side of things.

zstix:
TODO: add PR title and description (what parts of this PR should we look at?)

zstix:
I think that it would be best to close this, since it will continuously be out of date as we update / add quickstarts. We should recreate the PR just before we are ready to merge.

PR #791, opened 2022-02-04
Limit concurrent requests to nerdgraph

### LINKED ISSUE #777: `401` errors from Nerdgraph
## Description

The validation workflow that verifies quickstarts via the nerdgraph API is intermittently returning a `401` error. 

## Steps to Reproduce

- https://github.com/newrelic/newrelic-quickstarts/runs/5022458251?check_suite_focus=true

## Expected Behavior

- [ ] Success/failure of validation is reported.

### LINKED ISSUE #783: Quickstart validation hitting rate limiting
## Description
The new Nerdgraph validation for quickstarts is hitting API rate limits for large pull requests. 

## Steps to Reproduce
1. Have a PR with lots of changed quickstarts, such as #779 
2. Notice that the validation against nerdgraph starts to throw errors, [here](https://github.com/newrelic/newrelic-quickstarts/runs/5038388168?check_suite_focus=true)

## Expected Behavior
- [ ] The validation workflow is able to handle any number of quickstarts in a pull request.

aswanson-nr:
# Summary
* The concurrency of requests to nerdgraph during validation was
  overloading some systems and making us bad users of the api.
* This limits the number of concurrent requests per job to 5

## TODO
- [x] Add retries to calls to nerdgraph
- [ ] Maybe add more logging?

## Resolutions
- Closes #777 
- Closes #783

### DIFF
--- a/utils/nr-graphql-helpers.js
+++ b/utils/nr-graphql-helpers.js
@@ -120,8 +120,30 @@ const getCategoryTermsFromKeywords = (configKeywords = []) => {
   return categoryKeywords.length > 0 ? categoryKeywords : undefined;
 };
 
+/**
+ * Breaks an array up into parts, the last part may have less elements
+ * @param {Array} array - an array of anything
+ * @param {Number} chunkSize - the size of the parts
+ * @returns {Array} the array broken out into smaller array chunks
+ */
+const chunk = (array, chunkSize) => {
+  let chunkedArray = [];
+  let j = array.length;
+
+  if (chunkSize < 1) {
+    return chunkedArray;
+  }
+
+  for (let i = 0; i < j; i += chunkSize) {
+    chunkedArray = [...chunkedArray, array.slice(i, i + chunkSize)];
+  }
+
+  return chunkedArray;
+};
+
 module.exports = {
   fetchNRGraphqlResults,
   translateMutationErrors,
   getCategoryTermsFromKeywords,
+  chunk,
 };
--- a/utils/nr-graphql-helpers.js
+++ b/utils/nr-graphql-helpers.js
@@ -1,5 +1,6 @@
 'use strict';
 const fetch = require('node-fetch');
+const { Policy } = require('cockatiel');
 const instantObservabilityCategories = require('./instant-observability-categories.json');
 
 const NR_API_URL = process.env.NR_API_URL;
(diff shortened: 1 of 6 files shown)


utils/nr-graphql-helpers.js
```diff
@@ -25,17 +26,23 @@ const fetchNRGraphqlResults = async (queryBody) => {
   let results;
   let graphqlErrors = [];
 
+  // To help us ensure that the request hits and is processed by nerdgraph
+  // This will try the request 3 times, waiting a little longer between each attempt
+  const retry = Policy.handleAll().retry().attempts(3).exponential();
```
nr-kkenney, line 31: i think handleAll will only work for errors. for certain status codes, i think we will need something like:
```js
  const policy = Policy.handleWhenResult(
    (response) => response.status === 429 || response.status === 401
  )
    .retry()
    .attempts(3)
    .exponential();
```

nr-kkenney (changes requested):
looks pretty slick. dropped one comment, but other than that :+1:

utils/nr-graphql-helpers.js
aswanson-nr: Ah good point, we aren't actually throwing any errors here. Although I think we want to handle anything that isn't in the 200 range, right?
```suggestion
 const policy = Policy.handleWhenResult(
    (response) => response.statusCode < 200 || response.statusCode > 299
  )
    .retry()
    .attempts(3)
    .exponential();
```

aswanson-nr:
> another thing. i remember you ran into an issue with the response not returning correctly when using batches. has that been resolved? if not, does it need to be before merging?
That's still an issue with transport batching, I just defaulted back to running 5 concurrent requests instead.

utils/nr-graphql-helpers.js
nr-kkenney: we're probably fine to do that in general, probably even more fine with nerdgraph.

but if i think of something like 301, we wouldn't want to retry that. maybe anything 4**-5**.

aswanson-nr: I like that, will push a commit in a few.

aswanson-nr: @nr-kkenney Pushed!

nr-kkenney (approved):

PR #797, opened 2022-02-07
chore: icon to logo changes

jpvajda:
We currently don't use icons so this PR cleans up any misc quickstarts using icons. 

Note: we are going to be converting logos to icons after we switch 💯  over to the new service, but for now logos are the way....

To resolve` linkedEntityGuids` in 2 quickstarts I had to set some `linkedEntityGuids` to nulll


zstix (approved):
This looks good to me.

It seems like your editor might have formatted some of these files. Either way, things seem to be working well. Thanks!

PR #805, opened 2022-02-08
Serialize nerdlet state

aswanson-nr:
# Summary
* Nerdlet state for install plans must be string JSON, it was an object instead
* Update mutations with names for easier future debugging

### DIFF
--- a/utils/create-validate-install-plans.js
+++ b/utils/create-validate-install-plans.js
@@ -12,7 +12,7 @@ const {
 } = require('./nr-graphql-helpers');
 
 const INSTALL_PLAN_MUTATION = `# gql 
-mutation (
+mutation QuickstartRepoInstallPlanMutation (
   $description: String!
   $dryRun: Boolean
   $displayName: String!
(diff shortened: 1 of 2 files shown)


LizBaker (approved):

utils/create-validate-install-plans.js
```diff
@@ -137,6 +137,7 @@ const createValidateUpdateInstallPlan = async (installPlanFiles) => {
   for (const reqChunk of chunkedInstallPlanRequests) {
     const chunkRes = await Promise.all(
       reqChunk.map(async ({ variables, filePath }) => {
+        console.log(JSON.stringify(variables));
```
LizBaker, line 140: do we want to keep this log in here?

aswanson-nr: Whoops

aswanson-nr: Removed!

LizBaker (approved):

PR #820, opened 2022-02-10
Report job status instead of step

### LINKED ISSUE #819: [Repository] Quickstart schema validation does not report back on failure
## Description
The schema validation job in the `Validate quickstarts` workflow is not reporting failure back to the pull requests that trigger it. On success, it reports back just fine.
None of the other jobs in this workflow appear to have this issue.

## Steps to Reproduce
1. Make a change to quickstart that would make it invalid
2. Create a pull request
3. Wait for the automation to run
4. Notice that the pull request shows that it's green and good to merge
5. Check the `Actions` tab for the repository and notice that there was an error

## Expected Behavior
Failure is reported back to the pull request so the contributor knows that changes are needed

## Example
* [PR](https://github.com/newrelic/newrelic-quickstarts/pull/818)
* [Failed job](https://github.com/newrelic/newrelic-quickstarts/runs/5146603741?check_suite_focus=true)

aswanson-nr:
# Summary
* Step status only reports on the step itself and not the entire job.
  This meant that if a job failed before the tracked step it would never
  get reported back to the PR that triggered it.
* This updates the workflow to use the job status, which will report
  back failure if any steps fail.

## Issues
* Closes #819


nr-kkenney (changes requested):
looks good.
we'd probably want the same for validate_install_plans.yml.

aswanson-nr:
@nr-kkenney Good catch! Added that to the commit

PR #848, opened 2022-02-18
Validate Quickstart Dashboard Name Uniqueness

brammerl:
# Summary

This PR creates a script that validates a quickstart's dashboard's name is unique. 

Issue: https://newrelic.atlassian.net/jira/software/c/projects/DEVEX/boards/2045?modal=detail&selectedIssue=DEVEX-2297

## Pre merge checklist

- [x] Did you check your dashboard image quality? -  Do they look good?
- [x] Did you check that your alerts actually work?
- [x] Did you check your descriptive content for voice, tone, spelling and grammar errors?
- [x] Did you attach images of your dashboards to the PR so we can see them working?

### Screenshots

Attach images of any visual changes, such as a dashboard here.


zstix:
We should probably update the documentation in the repo to let contributors know that they need to have unique names for their dashboards.

utils/check_quickstart_uniqueness.js
```diff
@@ -62,11 +37,11 @@ const main = () => {
   const nameMatches = getMatchingNames(nameAndPaths);
   const idMatches = getMatchingIds(idsAndPaths);
 
-  if (nameMatches.length == 0 && idMatches.length == 0) {
+  if (nameMatches?.length == 0 && idMatches.length == 0) {
```
nr-kkenney, line 40: was the intent for this change to prevent exceptions?

it looks awkward because ?.length == 0 will be false i believe, and in that case i would expect getMatchingNames to return empty array, rather than a value with no .length.

nr-kkenney (changes requested):
can we add unit tests for check_dashboard_name_uniqueness and/or the helper methods used?

jpvajda:
I am working through several questions in slack with DevEx on this change, so we'll hold off on merging until we have some more clarifications

utils/check_quickstart_uniqueness.js
brammerl: Yeah so during a unit testing in check_quickstart_uniqueness.test.js sometimes the nameMatches array came back undefined so trying to access length on undefined was creating a reference error.

nr-kkenney: i think this is safe to remove / doesn't need to be added. looks like getMatchingNames would return an empty array when input is empty.

all the tests on your branch pass without the optional chain.

utils/check_quickstart_uniqueness.js
```diff
@@ -62,11 +37,11 @@ const main = () => {
...
     return console.log(`All quickstart names and ids are unique`);
   }
 
-  if (nameMatches.length > 0) {
+  if (nameMatches?.length > 0) {
```
nr-kkenney, line 44: same comment here.

nr-kkenney (approved):

PR #850, opened 2022-02-19
feat(prometheus): added dashboards for prometheus-remote-write

tclayton-newr:
# Summary

## Pre merge checklist

- [x] Did you check you NRQL syntax? - Does it work?
- [x] Did you check your dashboard image quality? -  Do they look good?
- [x] ~Did you check that your alerts actually work?~ N/A
- [x] ~Did you include an InstallPlan and Documentation reference?~ N/A
- [x] Did you check your descriptive content for voice, tone, spelling and grammar errors?
- [x] Did you attach images of your dashboards to the PR so we can see them working?

### Screenshots

Attach images of any visual changes, such as a dashboard here.
image
image

PR #861, opened 2022-02-24
fix(util): expand santization for linkedEntityGuids

tclayton-newr:
# Summary

While creating https://github.com/newrelic/newrelic-quickstarts/pull/850 I noticed that the sanitization script didn't replace sections such as 
```yaml
...
linkedEntityGuids: [
      "someGuidId"
]
...
```

This PR just expands the Regex to optionally include any line returns between `[` and `]`
I tested this out locally against my own dashboard.json file and using https://www.regexpal.com/
If there are test suites to run against the script I am happy to update those. 

> Deleted checklist section since this is not an update to a quickstart.


jpvajda:
Thanks for correcting the regex! @tclayton-newr  We appreciate that.

@aswanson-nr || @nr-kkenney 👋  do either you have any concerns with this PR?

nr-kkenney (approved):

Issue #862, opened 2022-02-24
Remedy newrelic-quickstarts workflow failures


jpvajda:
## Summary
A PR was merged and the subsequent release workflow failed to run successfully. Looks like the failure was due to another PR being merged at roughly the same time, and that causing a problem in the push step of the workflow.

## Acceptance Criteria
- [ ] Rerun the workflow, watch it succeed
- [ ] Add in a git pull before the commit step to reduce the window when this could happen

PR #863, opened 2022-02-24
update commit, push steps to reduce errors

nr-kkenney:
## Summary

The changes made were:
* first change was moving the commit step closer to the push step in
the overall workflow. Now, branch protection happens first, then commit,
then push. This should reduce the time between commit and push.
* second change was adding a pull command before changes are committed.
This should tighten the window of this step failing due a change
to main in main not being pulled, before we try to push local changes.
* the last change is grouping the echo command in the commit step with
the commit command. Now, echo setting the output should only happen when
we have changes to commit. If we dont have any changes, echo shouldnt
occur, and the push step should not run.

## Links
Relates to: #862

### DIFF
--- a/.github/workflows/pr-merged.yml
+++ b/.github/workflows/pr-merged.yml
@@ -36,15 +36,6 @@ jobs:
         id: generate-uuids
         run: cd utils && yarn generate-uuids
 
-      - name: Commit changes
-        id: commit-changes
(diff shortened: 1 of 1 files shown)


aswanson-nr (approved):
Assuming you tested that syntax with the curly brackets, looks good to me!

PR #886, opened 2022-03-02
Chore: update links for glassbox quickstart

jcountsNR:
# Summary

Glassbox completed an integration guide hosted on their site, so I'm changing the generic landing page to the guide.


install/third-party/glassbox/install.yml
```diff
@@ -11,4 +11,4 @@ target:
 install:
   mode: link
   destination:
-    url: https://www.glassbox.com/integrations/
+    url: https://support.glassbox.com/hc/en-gb/articles/4405777280401-How-to-Connect-New-Relic
```
jpvajda, line 14: Thanks for the contribution, but I just double checking the expected user experience on this link click,  it seems to be behind a login, which may be a bit confusing for a user. Is that what is expected?

Screen Shot 2022-03-03 at 4 50 02 PM

jpvajda (changes requested):

install/third-party/glassbox/install.yml
jcountsNR: I noticed that too. Let me double check.

jcountsNR: @jpvajda Glassbox is checking internally if they can host this file outside of the authentication wall. I think we could approve this one and if they move it later we could change it. Right now it's the only document that's actually useful for glassbox integration to NR.

jpvajda: ok. Thanks for confirming.

jpvajda (approved):

PR #890, opened 2022-03-02
add NR events to validate/publish workflows

aswanson-nr:
# Summary
* Add the node agent to validate/publish scripts
* Add custom event for install plans and quickstarts to allow for
  tracking in New relic one

### DIFF
--- a/utils/create-validate-install-plans.js
+++ b/utils/create-validate-install-plans.js
@@ -172,7 +172,7 @@ const main = async () => {
   const installPlanFiles = filterInstallPlans(files);
   const hasFailed = await createValidateUpdateInstallPlan(installPlanFiles);
 
-  newrelic.recordCustomEvent('validate_or_update_install_plans', {
+  newrelic.recordCustomEvent('ValidateOrUpdateInstallPlans', {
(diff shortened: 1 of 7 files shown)


utils/create_validate_pr_quickstarts.js
```diff
@@ -416,6 +417,11 @@ const main = async () => {
 
   const hasFailed = await createValidateUpdateQuickstarts(files);
 
+  newrelic.recordCustomEvent('validate_or_update_quickstarts', {
```
nr-kkenney, line 420: i think we can improve the name of the event.

Pascal case looks to be the standard for event names / types users can query from.

utils/create-validate-install-plans.js
```diff
@@ -170,6 +172,11 @@ const main = async () => {
   const installPlanFiles = filterInstallPlans(files);
   const hasFailed = await createValidateUpdateInstallPlan(installPlanFiles);
 
+  newrelic.recordCustomEvent('validate_or_update_install_plans', {
```
nr-kkenney, line 175: same comment as below.

nr-kkenney (approved):

Issue #633, opened 2021-12-10
[Repository] empty config.yml files should fail our repo checks


jpvajda:
PR reference  https://github.com/newrelic/newrelic-quickstarts/pull/614 

We had a case where a PR contained empty config files and we allowed the PR checks to pass. We should ensure a config file isn't empty and also contains the correct fields of data on a PR and fail a check so the user knows this isn't allowed.

 if a config file is missing required fields, a PR check fails and the user is notified a config file isn't complete. 

### Acceptance Criteria

- [ ] test if this actually in place already via PR https://github.com/newrelic/newrelic-quickstarts/pull/665/
- [ ] Review the feedback given on a failure case and evaluate if it's helpful. 
- [ ] Review comment from @jbeveland27 and consider our approach
- [ ] Review @jerelmiller's comments for fun. 😉 

### TBD

- [ ] update this [validation workflow ](https://github.com/newrelic/newrelic-quickstarts/blob/main/.github/workflows/validate_packs.yml)to  to check to just run a dry run mutation so we can get back validation errors and then can notify a user their PR is missing required fields.

jpvajda:
I will follow up with DevEx if they are checking for this at the API level.

jpvajda:
From @jbeveland27  

_The PR workflow still needs a bit of lightweight validation on its own to make sure enough of the quickstart is present to construct that mutation input. all the required fields in the mutation must be provided, otherwise the mutation call won't even reach the catalog service. I think the validation workflow needs to be revamped in general. Schema checking doesn't need to occur (i.e. are required fields present), but that data needs to get wrapped up in a mutation call and then the response needs to be inspected. If you don't set the input correctly it doesn't matter what the mutation response is, it'll never get to the service because NerdGraph will just reject it._

jerelmiller:
This seems to me something that should be done repo-side. Keep in mind that we are designing the NerdGraph API in a way that facilitates data from everywhere. Using non-null fields is a way to communicate on our end that a field is required. Empty configuration files should be easy to detect repo-side without needing to run the GraphQL mutation to figure out its invalid.

The only other situation I could see being an issue is having a configuration file that omits non-null fields from the GraphQL mutation. The mutation will error out before it hits our validation because you'd be sending a `null` for a non-null field. There are a couple different ways I could think of to handle this:

1) Parse the error message and pull out the field names.

[...]

PR #935, opened 2022-03-17
chore: empty config test [dont merge]

jpvajda:
this a test to see how we handle an empty config related to this issue:

https://github.com/newrelic/newrelic-quickstarts/issues/633


quickstarts/empty-config-test/config.yml
```diff
@@ -0,0 +1 @@
+# this is an empty file
```
aswanson-nr, line 1:
```suggestion
```

aswanson-nr:
@jpvajda I deleted the comment in the `config.yml` just to be sure

jpvajda:
@aswanson-nr the jobs aren't running, and just queued, but I think that is related to the Github issues we've been seeing all day.

jpvajda:
@aswanson-nr nice! It seems we now have a very clear error in this case for a empty config, so I think we can  close out #633 unless you disagree. 

```
 Error: quickstarts/empty-config-test/config.yml
	 '#/type' must be object
```

zstix:
We have confirmed what was necessary for this PR. Will close this.

PR #958, opened 2022-03-25
Revert "Misc quickstart fixes"

zstix:
Reverts newrelic/newrelic-quickstarts#941

This should return the titles / names to their previous state.


jpvajda:
The change we need for a quick fix to gigamon should be covered in https://github.com/newrelic/newrelic-quickstarts/pull/959  but we can leave this open in case need to do a full revert

PR #963, opened 2022-03-25
Revert "Netlify fixz"

nr-kkenney:
Reverts newrelic/newrelic-quickstarts#932


jpvajda (approved):

jpvajda:
This is ok to approve but these should really be in the[ dashboard directory ](https://github.com/newrelic/newrelic-quickstarts/tree/main/quickstarts/netlify/dashboards) for clarity, also this will actually cause the image count error to fire again on this quickstart, as it's set to 6 total images so we'll need to up that limit.

PR #975, opened 2022-03-29
Remove redundant validation

josephgregoryii:


josephgregoryii:
`Validation / Quickstart schema compliance` fails in check, however my PR removes that check entirely

utils/generate-uuids.js
```diff
@@ -22,7 +22,7 @@ const main = (paths) => {
 };
 
 /**
- * This allows us to check if the script was invoked directly from the command line, i.e 'node validate_quickstarts.js', or if it was imported.
+ * This allows us to check if the script was invoked directly from the command line, i.e 'node geberate-uuids.js', or if it was imported.
```
aswanson-nr, line 25:
```suggestion
 * This allows us to check if the script was invoked directly from the command line, i.e 'node generate-uuids.js', or if it was imported.
```

aswanson-nr (changes requested):
Looks good! Could you also remove the `schemas` directory as well, since we no longer need them?

aswanson-nr (changes requested):
Looks good to me! But I just realized we'll need to update the branch protection rules so that they don't expect these jobs to exist. We'll need to update any workflows that disable/enable branch protection and update the settings once this is merged.

nr-kkenney (approved):

Issue #930, opened 2022-03-17
Publishing quickstart alert policy is not firing


aswanson-nr:
## Description
The alert policy for the quickstart submission workflow does not appear to be properly firing when a workflow fails.

## Helpful links
* [troubleshooting dashboard](https://staging.onenr.io/0Z2R58znkwb)
* [alert policy](https://staging.onenr.io/0DvwBAz2bQp)

## Steps to Reproduce
* Review https://github.com/newrelic/newrelic-quickstarts/actions/runs/1995821133/attempts/1
* Notice that there are no alerts in the bots channel for that time range

## Expected Behavior
An alert condition fires off and is visible within the #dev-enablement-bots channel

nr-kkenney:
okayyyyy. some updates. i was initially really confused 🙈  we have this block of code: 

```javascript
const hasFailed = await createValidateUpdateQuickstarts(files);
recordCustomNREvent(hasFailed, isDryRun === 'true');

if (hasFailed) {
  process.exit(1);
}
```

if we look at the erroring workflow, we know we are hitting the process.exit(1) because the step fails. that means `hasFailed` is `true`. heres what really confused me -- if `hasFailed=true`, then the event we are sending in recordCustomNREvent is correct.  so everything is correct, we're just not sending the event. 

if we look in the newrelic UI, we are sending quite a few events -- but all of them are successes, no failures. this is the same for validations vs update mutations.

so to recap:
* code looks correct at every step, but we are not seeing failure events
* in the UI we see no failure events period -- even for validation.

so we see no events for failures, but it looks like it should be sending them. i dont actually know what is mechanically happening under the hood for the newrelic agent, but I had a hunch that because we exit the program nearly immediately after trying to send a custom event, that perhaps the agent wasn't able to send it -- similar to if the program crashes. because this is a github workflow, there isn't an opportunity for the agent to restart and pick up where it left off like it might for a regular apm application.

turns out if i add enough delay to the app as it runs before we exit, we actually send & can see the event in newrelic.

PR #982, opened 2022-03-30
fix for custom events & alerting on failures

nr-kkenney:
# Summary

Turns out the code is structured correctly, but the agent wasn't sending events. The call to `newrelic.recordCustomEvent` is just queueing up the event to be sent, and the agent is asynchronously sending data to new relic. When we hit `process.exit(1)`, it seems to be the case that the event hasn't been sent yet, and the agent wont be able to send it at all.

Solution:
Swapped from using the agent to using the api to record custom events.

Resolves: #930


utils/create-validate-install-plans.js
```diff
@@ -164,16 +164,16 @@ const createValidateUpdateInstallPlan = async (installPlanFiles) => {
 
 /**
  * @param {boolean} hasFailed if the validation or submission has failed
- * @param {boolean} isDryRun - true for validation, false for submission
+ * @param {Promise<boolean>} isDryRun - true for validation, false for submission
```
nr-kkenney, line 167: that should be return 🙈  whoops

utils/newrelic/customEvent.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * This module is intended for use only with SSR pages.
+ * For monitoring & creating events for browser pages, use the browser agent instead.
+ */
```
nr-kkenney, line 4: dont need these comments

utils/create-validate-install-plans.js
```diff
@@ -164,16 +164,16 @@ const createValidateUpdateInstallPlan = async (installPlanFiles) => {
...
  */
-const recordCustomNREvent = (hasFailed, isDryRun) => {
+const recordCustomNREvent = async (hasFailed, isDryRun) => {
   const status = hasFailed ? 'failed' : 'success';
   if (isDryRun) {
-    newrelic.recordCustomEvent('ValidateInstallPlans', {
+    await track('ValidateQuickstarts', {
```
nr-kkenney, line 172: these should be install plan events

josephgregoryii (approved):
looks Grrrreat!

josephgregoryii (approved):

PR #990, opened 2022-04-04
Docs-based install plans for mobile quickstarts

cjmccarthy:
# Summary

Mobile quickstarts don't have install plans, which makes them impossible to compose with other quickstarts into a combined install plan. This PR converts the Mobile quickstarts to function exactly like the Browser quickstart, with install plans that point to the docs. This will allow us to create multi-step install plans for quickstarts that require the mobile agent to be in place first, similar to what we did for our browser-based video trackers ([example](https://newrelic.com/instant-observability/video-web-videojs/e0296927-4ea9-40a8-a1d4-3a2c1cced9af)).

A fast follow to this PR will be another PR with mobile video tracker quickstarts utilizing these new install plans, similar to the example above.


install/mobile/android/install.yml
```diff
@@ -0,0 +1,14 @@
+id: android-docs
```
jpvajda, line 1: I'm curious why this ID is using the term "docs" in it? The intention would be to use this install plan for other installation cases, besides just viewing docs links, so I'd suggest to name this more declarative  like `android-mobile`

install/mobile/ios/install.yml
```diff
@@ -0,0 +1,14 @@
+id: ios-docs
```
jpvajda, line 1: same suggestion in this ID I'd suggest  `ios-mobile`

install/mobile/tvos/install.yml
```diff
@@ -0,0 +1,14 @@
+id: tvos-docs
```
jpvajda, line 1: same suggestion: `tvos-mobile`

jpvajda (changes requested):
Thanks for the contribution, I just had some suggestions!

install/mobile/android/install.yml
cjmccarthy: I simply copied this from the browser quickstart, which used this -docs convention. Will change per your suggestion

PR #997, opened 2022-04-06
Name fix for slug

josephgregoryii:
## Summary

This PR is for fixing any quickstarts that had a `name` field in `config.yml` that would end up breaking the quickstart's full URL if the quickstart got updated. 

## Process of the script:

1. Find all quickstarts that have a different `name` and `title` field
2. find all quickstarts that havae the same `name` and `title` field.

For items in `1.` - slugify the title and set it to name IF the current name leads to an HTTP request returning a 404. Otherwise, keep the name as-is

for items in `2.` - Ensure the current `name` field is correct and slugified. URL Should not return a 404


utils/check-quickstart-slug.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  const fetchResults: Promise<string>[] = diffs.map(async (tuple) => {
+    const [quickstartPath, contents] = tuple;
+    const resp = await fetch(`${SITE_URL}/${contents.name}/${contents.id}`);
+    if (!resp.ok || resp.status === 404) {
+      return quickstartPath;
+    }
+  });
+  return Promise.all(fetchResults);
+};
```
josephgregoryii, line 91: This function is for verifying `contents.name` is the correct URL for the quickstart. (soon to be `contents.slug`)

josephgregoryii: Also to note: I left the script in the PR for code review purposes. We do not have to check it into the repo's `main`

nr-kkenney (dismissed):

nr-kkenney (approved):

PR #1000, opened 2022-04-06
Content change name to slug

josephgregoryii:


quickstarts/apis/event/config.yml
```diff
@@ -1,5 +1,5 @@
 id: bc62e8f8-0d0c-4c86-9f05-b9e83e21d0fc
-name: event-api
+slug: Event API
```
aswanson-nr, line 2: Looks like this one went the opposite direction!

nr-kkenney (dismissed):

aswanson-nr (commented):
Looking good! Just found a few slugs that need to be updated

aswanson-nr (approved):

Issue #934, opened 2022-03-02
[Repository] Explore changes to quickstart submission


nr-kkenney:
## Summary
We ran into a few failures documented [here](https://github.com/newrelic/newrelic-quickstarts/issues/881). One of the observations we made was that when a workflow fails to submit quickstarts for the staging environment, production and EU are both skipped.

Additionally, if we fail for any environment, when we need to rerun the workflow, all environments are included. So we fail for the EU environment, we rerun EU, staging, production -- but we just want to rerun EU.

We want to try answer the following questions:
- should quickstarts be submitted across all environments in parallel? is there a reason to go to staging first, then maybe EU, prod in parallel?
- can we set up the workflow(s) so that these are separate, with the ability to pass/fail independently of another environment?

## Acceptance Criteria
- [x] set up the workflow(s) so that these are separate, with the ability to pass/fail independently of another environment
- [x] a workflow can be manually run for each environment as needed.

jerelmiller:
I could see prod and EU being run in parallel. One advantage of running staging first is that you can catch any issues you may see before it hits prod/EU. That may or may not be enough of a reason to not make that parallel as well, but wanted to throw that out there as a potential reason to keep staging first.

nr-kkenney:
ive been testing out most of my changes on a fork / personal repository. i now have the workflow(s) set up, and the summary looks like:

![Screen Shot 2022-04-06 at 4 54 21 PM](https://user-images.githubusercontent.com/70179215/162093497-a5e005f6-9112-4cc6-a1ab-c7445a2f776a.png)

a slight bit of testing is still needed, but should be able to get a PR out tomorrow.

nr-kkenney:
With how i set things up, having the validation workflows use the reusable workflow would be a bit messy. The reusable workflow does both install plan & quickstart stuff in one run. The validation we currently have in the repo is separate for install plans & quickstarts, with no clean way to use the reusable workflow.

Additionally, the validation workflows actually have more jobs than the reusable workflow does -- they are validating multiple things, sometimes using many scripts. The reusable workflow only validates against the API (schema validation).

I could setup two reusable workflows — one for install plans, one for quickstarts — but i actually think we dont want this, since it might be on the table to actually just have one entrypoint script in the future — the CLI.

PR #1008, opened 2022-04-08
add reusable submission workflow, submission workflows for each environment

nr-kkenney:
## Summary
This PR adds in a new reusable workflow which can validate or submit install plans & quickstarts for a given environment. Additionally, a new workflow has been added for each environment. This allows us to run changes for only a single environment as needed.

Finally, the core submission workflow has been updated to utilize this new reusable workflow. With this change, the overall process for when a pr is merged is now slightly different. Staging runs before the other environments, with production & eu-production running simultaneously. For each environment, install plans are first added, then quickstarts. 

Previously, the process was:
 1. staging install plans added.
 2. production install plans added.
 3. eu-production install plans added.
 4. staging quickstarts added.
 5. production quickstarts added.
 6. eu-production quickstarts added.

## Screenshots
Here is an example run summary with the new setup: 
Screen Shot 2022-04-06 at 4 54 21 PM

## Links
Resolves: #934


.github/workflows/pr-merged.yml
```diff
@@ -133,97 +134,35 @@ jobs:
...
-          NR_API_TOKEN: ${{ secrets.NR_API_TOKEN_EU }}
-        run: |
-          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ steps.get_pr_number.outputs.pr-number }}/files"
-          cd utils && yarn create-validate-pr-quickstarts $URL ${DRY_RUN}
+  staging:
+    needs: [generate-uuid, get-pr-number]
+    uses: ./.github/workflows/reusable.quickstart_submission.yml
+    with:
+      pr-number: ${{ needs.get-pr-number.outputs.pr-number }}
+      dry-run: true
```
nr-kkenney, line 142: these should be false for real submission 🙈

.github/workflows/quickstarts.eu-production.yml
```diff
@@ -0,0 +1,19 @@
...
+      pr-number:
+        description: 'PR to run against'
+        required: true
+
+jobs:
+  eu-production:
+    uses: ./.github/workflows/reusable.quickstart_submission.yml
+    with:
+      pr-number: ${{ github.event.inputs.pr-number }}
+      dry-run: true
```
nr-kkenney, line 15: same comment as above. should be false for real submissions.

aswanson-nr (commented):
Clean!! this looks great! 
Can we use this reusable workflow for the validation as well?
Also I remember talking about having the manual publishing workflows take in the quickstart ID instead of working with a PR, but that would definitely require changes on the script side, so this looks fantastic for the first pass!

nr-kkenney:
I'll also copy this comment over to the ticket.

With how i set things up, having the validation workflows use the reusable workflow would be a bit messy. The reusable workflow does both install plan & quickstart stuff in one run. The validation we currently have in the repo is separate for install plans & quickstarts, with no clean way to use the reusable workflow. Additionally, the validation workflows actually have more jobs than the reusable workflow does -- they are validating multiple things, sometimes using many scripts. The reusable workflow only validates against the API (schema validation).

i could setup two reusable workflows — one for install plans, one for quickstarts — but i actually think we dont want this, since it might be on the table to actually just have one entrypoint script in the future — the CLI.

aswanson-nr (approved):

PR #1010, opened 2022-04-11
Update coredns-panics.yml

lazyplatypus:
# Summary
It seems like there is a syntax issue with the alert - hopefully this doesn't mess up the mutation.


lazyplatypus:
@tclayton-newr

aswanson-nr (approved):

PR #1014, opened 2022-04-11
fix: changes to fix third party notices workflow job

nr-kkenney:
## Summary

* have all the workflow steps run in the utils directory, since our
code doesn't live at the root of the repo.
* add a third_party_manifest.json file.

## Links

workflow failure: https://github.com/newrelic/newrelic-quickstarts/runs/5979879164?check_suite_focus=true


josephgregoryii (approved):

aswanson-nr:
I'm just a little worried that the third party notices won't work under the `utils` directory and expects it to be at the top level. Just a thought I had, it probably works just fine.

aswanson-nr (approved):

PR #1017, opened 2022-04-12
Fix NRQL parsing in sanitize dashboard script

aswanson-nr:
# Summary
We had an issue get through for a newly created dashboard, where the sanitize-dashboard script was removing commas within NRQL. It is only supposed to remove trailing slashes from the dashboard JSON. Looks like the regex for removing trailing commas is picking up the `),` in the NRQL. I tested removing `\)` from this [regex]( https://github.com/newrelic/newrelic-quickstarts/blob/main/utils/sanitize_dashboards.js#L16) and that seemed to do the trick. I can’t think of any situations where we would need to remove a trailing comma after a closing parentheses in json, so this seems like a change worth making.


aswanson-nr:
https://github.com/newrelic/newrelic-quickstarts/pull/1016/files is the PR fixing the issue, as you can see there were curly brackets in the NRQL instead of the closing parentheses and comma.

josephgregoryii (approved):

PR #1020, opened 2022-04-12
Script, tests, and workflow updates for icon and description fields

josephgregoryii:
### Summary

This PR is for the code changes inside our `utils/` directory and workflow updates. Alert config files now have `description` which replaces `details` to better align with the catalog service.


utils/check_dashboard_name_uniqueness.ts
```diff
@@ -23,8 +23,9 @@ interface DashboardConfigs {
 /** Returns any quickstart dashboards with matching names and their filepaths */
 const findMatchingDashboardNames = (): DashboardNamesAndPaths[] => {
   const dashboardConfigs = glob.sync(
-    path.resolve(process.cwd(), '../quickstarts/**/dashboards/*/*.+(json)')
+    path.resolve(process.cwd(), '../quickstarts/**/dashboards/*.+(json)')
```
aswanson-nr, line 26: We want to keep this glob pattern the same as main. Merging in the main branch should get you there

.github/workflows/quickstarts.eu-production.yml
```diff
@@ -0,0 +1,19 @@
+name: Submit All - EU-Production
```
josephgregoryii, line 1: im not sure why these are showing up as new files when i pulled it from main... we can probably ignore these workflow files

nr-kkenney (approved):

PR #1027, opened 2022-04-13
Image validation fix

josephgregoryii:
## Summary

This PR adds an additional step in logic to confirm that the number of dashboards is greater than 0. If `dashboardCount` is `0`, the if statement would hit an incorrect truthy statement.
https://github.com/newrelic/newrelic-quickstarts/blob/2d60bc77b5e5963e6d22cf7700dfb827343911e8/utils/validate_images.js#L56

Above, no matter what the value of `screenshotPaths.length` is, it would equal `true` if `dashboardCount` is `0`

Update:
https://github.com/newrelic/newrelic-quickstarts/blob/fc50d87e4ce23ea7ffb5db75a88ff118cf5be206/utils/validate_images.js#L56

### DIFF
--- a/utils/validate_images.js
+++ b/utils/validate_images.js
@@ -53,7 +53,7 @@ const validateImageCounts = (quickstartDirs) => {
     );
 
     // Each dashboard is allowed MAX_NUM_IMG dashboards
-    if (screenshotPaths.length > MAX_NUM_IMG * dashboardCount) {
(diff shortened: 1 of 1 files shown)


aswanson-nr:
@josephgregoryii Do you mean that it would hit an incorrect falsey statement? Since `0` is typecast to false.

PR #1028, opened 2022-04-13
Create initial two Netlify data sources

zstix:
## Description

- Created a top-level `data-sources` directory
- Crated two data sources for Netlify (builds and logs)

We populated the `config.yml` files with the following content:

```yml
# human created
id: netlify-builds

# Display name from the install plan and the quickstart
displayName: Netlify Builds

# Description from the install plan
description: |
  Netlify is an all-in-one platform for automating modern web projects. Replace your hosting infrastructure, continuous integration, and deployment pipeline with a single workflow. Integrate dynamic functionality like serverless functions, user authentication, and form handling as your projects grow.

install:
  primary:
    # always going to be quickstart install nerdlet (see ticket)
    nerdlet: nr1-install-newrelic.quickstart-installation-plan
    nerdlet_state:
      # ID for the related quickstart
      quickstart_id: 5861d7f5-26c0-43ad-bda8-c893c4b27b25
      requires_account: true
  secondary:
    # install link from the install plan
    link: https://www.npmjs.com/package/@newrelic/netlify-plugin#installation

# keywords from the quickstart
keywords:
  - infrastructure
  - automation
  - Netlify
  - Jamstack
  - RUM
  - browser agent
  - continuous integration
  - continuous deployment
  - newrelic partner

# hard-coded, always partner (for these first initial data sources)
categoryTerms:
  - newrelic partner
```

I decided to remove `logo: logo.png` since it seems like it's always going to be the same thing. We'll just need to look up the image in the data-source directory and use that for submitting the new data-source.

## Related Issue(s)

- NR-836


zstix:
We _might_ want to merge this into a feature branch so we can test our workflows for validating / submitting into `main`.

data-sources/netlify-builds/config.yml
```diff
@@ -0,0 +1,30 @@
...
+  - infrastructure
+  - automation
+  - Netlify
+  - Jamstack
+  - RUM
+  - browser agent
+  - continuous integration
+  - continuous deployment
+  - newrelic partner
+  - featured
```
jerelmiller, line 27: [nit] There is no concept of "featured" for data sources, so this keyword might not make much sense here anymore. All this effects is the results returned when doing a keyword search, so no harm in leaving it, but just letting you know its not needed for the "fetaured" functionality in I/O.

zstix: Easy enough to remove!

data-sources/netlify-builds/config.yml
```diff
@@ -0,0 +1,29 @@
...
+displayName: Netlify Builds
+
+description: |
+  Netlify is an all-in-one platform for automating modern web projects. Replace your hosting infrastructure, continuous integration, and deployment pipeline with a single workflow. Integrate dynamic functionality like serverless functions, user authentication, and form handling as your projects grow.
+
+install:
+  primary:
+    nerdlet: nr1-install-newrelic.quickstart-installation-plan
+    nerdlet_state:
+      quickstart_id: 5861d7f5-26c0-43ad-bda8-c893c4b27b25
```
aswanson-nr, line 12: I may have missed this while we were discussing the ticket, is this referencing back to the quickstart so that we can associate dashboards and alerts before we have the ability to directly reference connect them to the datasource?

aswanson-nr (commented):
Overall looks good! Just had a question around the quickstart id being included here.

data-sources/netlify-builds/config.yml
jerelmiller: Correct! As a workaround for now, we are planning to have some of the data sources just send you through the quickstart install flow, hence the need for the associated quickstart ID.

zstix: Yep!

This was something that Jerel and Jewels suggested as a short-term workaround. When a user goes to install a datasource, it will drop them into the install UI for the _quickstart_ so that it also installs the desired dashboards and alerts.

I tried to add some details to the ticket...but it was late in the day yesterday (well after the team had originally looked at and estimated it).

aswanson-nr: Makes total sense! Thanks for the clarification.

aswanson-nr (approved):

PR #1029, opened 2022-04-14
add yaml validation

nr-kkenney:
## Summary

* added new workflow which runs yamllint on push
* added yamllint config file

## Example Output
Screen Shot 2022-04-14 at 9 57 54 AM

### DIFF
--- a/.config/yamllint/config.yml
+++ b/.config/yamllint/config.yml
@@ -4,6 +4,7 @@ extends: relaxed
 
 ignore: |
   **/node_modules/
+  **/utils/__tests__/mock_files/
 
 rules:
   colons: disable
--- a/.config/yamllint/config.yml
+++ b/.config/yamllint/config.yml
@@ -1,0 +1,14 @@
+---
+
+extends: relaxed
+
+ignore: |
+  **/node_modules/
+
+rules:
+  colons: disable
+  empty-lines: disable
+  indentation: disable
+  line-length: disable
+  new-line-at-end-of-file: disable
+  trailing-spaces: disable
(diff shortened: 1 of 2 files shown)


.github/workflows/yaml-lint.yml
```diff
@@ -0,0 +1,18 @@
+name: Lint Yaml
+
+on:
+  push:
+    paths:
+      - '**/*.yml'
```
nr-kkenney, line 6: probably also want .yaml

nr-kkenney: also looks like the pattern can just be **.yml, **.yaml

.github/workflows/yaml-lint.yml
```diff
@@ -0,0 +1,19 @@
+name: Lint Yaml
+
+on:
+  push:
```
nr-kkenney, line 4: should be `pull_request` i think, if a commit doesn't have yml files but the PR in general does.

.config/yamllint/config.yml
```diff
@@ -0,0 +1,14 @@
+---
+
+extends: relaxed
+
+ignore: |
+  **/node_modules/
```
nr-kkenney, line 6: might want to also exclude testing folder mock_files

.github/workflows/yaml-lint.yml
nr-kkenney: pull_request should also work with forks, since we don't need secrets or write permissions to run validation

aswanson-nr (approved):

PR #1033, opened 2022-04-18
Feature/typescript utils

aswanson-nr:
# Summary

Converts helper functions and the validation/submission of quickstarts and install plans over to typescript


utils/package.json
```diff
@@ -5,23 +5,23 @@
...
+    "validate-icons": "ts-node validate_icons.js",
+    "generate-uuids": "ts-node generate-uuids.js",
+    "sanitize-dashboard": "ts-node sanitize_dashboards.js",
+    "generate-schema-docs": "ts-node graphql-schemas/generate-schema-docs.js"
   },
   "dependencies": {
     "@actions/core": "^1.4.0",
     "@types/glob": "^7.2.0",
-    "@types/node": "^17.0.23",
+    "@types/node": "^17.0.24",
```
zstix, line 24: It looks like we have _some_ types in `devDependencies` and _some_ in `dependencies`. Because we don't actually bundle this stuff for production, we can probably just put all of these in dependencies (or dev dependencies...your call).

utils/types/nerdgraph.ts
```diff
@@ -0,0 +1,52 @@
...
+export interface NerdGraphRequest<T = {}> {
+  queryString: string;
+  variables?: T;
+}
+
+export interface QuickstartMutationResponse {
+  quickstart: {
+    id: string;
+  };
+}
```
zstix, line 52: This could probably be co-located with the code that uses it.

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
+export interface QuickstartAlertInput {
+  description?: string;
+  displayName: string;
+  rawConfiguration: string;
+  sourceUrl?: string;
+  type: string;
```
zstix, line 6: Could this be an enum?

zstix:
```suggestion
  type: 'BASELINE' | 'STATIC';
```

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
...
+  categoryTerms?: string[];
+  description: string;
+  displayName: string;
+  slug?: string;
+  documentation?: QuickstartDocumentation[];
+  icon: string;
+  keywords?: string[];
+  sourceUrl?: string;
+  summary: string;
+  supportLevel?: string;
```
zstix, line 43: These could probably be set in the type file:

```suggestion
  supportLevel?: 'COMMUNITY' | 'ENTERPRISE' | 'NEW_RELIC' | 'VERIFIED';
```

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
...
+  dashboards?: QuickstartDashboardInput[];
+}
+
+export interface QuickstartMutationVariable {
+  id: string;
+  dryRun: boolean;
+  quickstartMetadata: QuickstartMetaData;
+}
+
+// --------------------- YML / CONFIG Types
```
zstix, line 54: Since the name of this file signifies that this is for _MutationVariables_ we might want to either rename this or move the YAML stuff to it's own file.

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
...
+
+// or AlertConfig, etc
+export interface AlertYml {
```
zstix, line 57: Let's go with the suggested name change:

```suggestion
export interface AlertConfig {
```

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
...
+  name: string;
+  description: string;
+  type: string;
```
zstix, line 60: We might want to make a type for this:

```ts
type AlertType = 'BASELINE' | 'STATIC';
```

Then we can use that type here and in the variables interface above.

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export interface QuickstartDocumentationYml {
```
zstix, line 63: If we decide to remove `Yml` from the other interface, we should probably remove it here too.

utils/types/QuickstartMutationVariable.ts
```diff
@@ -0,0 +1,81 @@
...
+  description: string;
+  title: string;
+  slug?: string;
+  documentation: QuickstartDocumentationYml[];
+  icon: string;
+  keywords?: string[];
+  summary: string;
+  installPlans?: string[];
+  id: string;
+  level: string;
```
zstix, line 80: Like alerts, we can probably make a type for level and use it in both the variables and the YAML config interfaces. Something like `QuickstartLevel`?

utils/types/InstallPlanMutationVariables.ts
```diff
@@ -0,0 +1,59 @@
...
+interface LinkDirective {
+  link: {
+    url: string;
+  };
+}
+
+interface NerdletDirective {
+  nerdlet: {
+    nerdletId: string;
+    nerdletState: string;
```
zstix, line 39: It looks like this is optional:

```suggestion
    nerdletState?: string;
```

utils/types/InstallPlanMutationVariables.ts
```diff
@@ -0,0 +1,59 @@
...
+  };
+}
+
+interface DefaultDirective {
+  mode: string;
+  destination: undefined;
```
zstix, line 45:
```suggestion
  destination?: string;
```

zstix (commented):
WIP Review

utils/types/InstallPlanMutationVariables.ts
aswanson-nr: We need to determine whether we can actually hit the case where this type would be used. I will chat with Devex

utils/package.json
```diff
@@ -5,23 +5,23 @@
   "author": "NewRelic",
   "license": "Apache-2.0",
   "scripts": {
-    "validate-quickstart-install-plans": "node validate-quickstart-install-plans.js",
-    "create-validate-install-plans": "node create-validate-install-plans.js",
-    "create-validate-pr-quickstarts": "node create_validate_pr_quickstarts.js",
+    "validate-quickstart-install-plans": "ts-node validate-quickstart-install-plans.js",
```
nr-kkenney, line 8: these might have to be `npx ts-node`. depends on how these are installed i think.

aswanson-nr (commented):
Looks like there a still a few things to clean up and one or two discussions to have

utils/package.json
nr-kkenney: moved into dependencies

utils/types/nerdgraph.ts
nr-kkenney: moved

utils/types/QuickstartMutationVariable.ts
nr-kkenney: moved

utils/create-validate-install-plans.ts
```diff
@@ -1,70 +1,100 @@
-'use strict';
```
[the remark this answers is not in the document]
aswanson-nr: Added the `strict` option to the TSconfig as a replacement

aswanson-nr:
Install plan run https://github.com/nr-kkenney/newrelic-quickstarts/runs/6069791055?check_suite_focus=true
Quickstart run https://github.com/nr-kkenney/newrelic-quickstarts/runs/6069791409?check_suite_focus=true

We expected both of these to fail, since we introduced incorrect changes. Both ran as expect and this should be good to merge.

utils/create_validate_pr_quickstarts.ts
```diff
@@ -252,9 +282,14 @@ const getQuickstartDashboardConfigs = (quickstartConfigPath) => {
...
-const adaptQuickstartDashboardInput = (dashboardConfigPaths) =>
+const adaptQuickstartDashboardInput = (
+  dashboardConfigPaths: string[]
+): QuickstartDashboardInput[] =>
   dashboardConfigPaths.map((dashboardConfigPath) => {
-    const parsedConfig = readQuickstartFile(dashboardConfigPath);
+    const parsedConfig = readQuickstartFile<{
+      description: string;
+      name: string;
+    }>(dashboardConfigPath);
```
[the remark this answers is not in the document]
aswanson-nr: Since these are the only fields we care about on dashboards and the rest is just treated as a string, I think this type is just fine.

utils/create-validate-install-plans.ts
```diff
@@ -1,70 +1,100 @@
...
-      id: $id
-      primary: $primary
-      target: $target
-    }
+} from './nr-graphql-helpers';
+import { track, CUSTOM_EVENT } from './newrelic/customEvent';
+
+const gql = String.raw;
+export const INSTALL_PLAN_MUTATION = gql`
+  # gql
```
zstix, line 31: Do we need this?

utils/create-validate-install-plans.ts
```diff
@@ -1,70 +1,100 @@
...
+interface InstallPlanMutationResponse {
+  installPlanStep: {
+    id: string;
+  };
+}
+
 /**
  * Builds the target parameter from the config into the variables for NR Request
  * @param {Object} target the `target` parameter object
  * @returns {Object} target transformed for NR request
```
zstix, line 70: We probably don't want to have types defined in two places. It would be a bummer if an IDE read from these types instead of the actual Typescript types.

zstix (approved):

zstix (approved):

utils/create-validate-install-plans.ts
```diff
@@ -1,70 +1,105 @@
...
-const {
+import {
+  InstallPlanDestination,
+  InstallPlanDirectiveInput,
+  InstallPlanMutationVariable,
+  InstallPlanOperatingSystem,
+  InstallPlanTargetInput,
+  InstallPlanTargetType,
+} from './types/InstallPlanMutationVariables';
+import {
```
josephgregoryii, line 9: Super nitpick - Personally I think its easier to read `import type {...} from '...'` when it comes to importing types (especially sinbce this file has so many), however I think this is something we can achieve throughout converting to TS.

josephgregoryii (approved):

PR #1040, opened 2022-04-20
add more data sources

nr-kkenney:
## Summary

High level: adding in remaining data sources & logos.

* id is just the general name of the data source.
* displayName is copied from the title of their respective quickstarts.
* The logos were copied from their respective quickstarts. 
* Descriptions were also copied from respective quickstart.
* The fallback/secondary link is the first documentation url from the quickstart, or if one didn't exist, just a reasonable link from the description or summary of the quickstart.
* Keywords are copied from the quickstart and should be identical between the two.
* CategoryTerms are my manual input from knowing how keywords & categories work. 
  * As an example, if a quickstart had `infrastructure` `newrelic partner` keywords, i put `infrastructure` and `newrelic partner` as category terms because i know those SHOULD map correctly and preserve the categories of the quickstart to the data source.


data-sources/atlassian-jira/config.yml
```diff
@@ -0,0 +1,31 @@
...
+logo: logo.png
+
+install:
+  primary:
+    nerdlet: nr1-install-newrelic.quickstart-installation-plan
+    nerdletState:
+      quickstart_id: a99b22a5-5666-4096-b239-8f315700fde0
+  requiresAccount: true
+  secondary:
+    link: https://docs.newrelic.com/docs/alerts-applied-intelligence/notifications/notification-integrations/#jira
```
zstix, line 19:
```suggestion
install:
  primary:
    nerdlet: nr1-install-newrelic.quickstart-installation-plan
    nerdletState:
      quickstart_id: a99b22a5-5666-4096-b239-8f315700fde0
    requiresAccount: true
  secondary:
    link: https://docs.newrelic.com/docs/alerts-applied-intelligence/notifications/notification-integrations/#jira
```

data-sources/atlassian-jira/config.yml
```diff
@@ -0,0 +1,31 @@
+id: atlassian-jira
+
+displayName: Atlassian JIRA
+
+description: |
+  Jira is a proprietary issue tracking product developed by Atlassian that allows bug tracking and agile project management. You can Integrate New Relic with Atlassian Jira(Cloud) and automatically create and update Jira issues.
+
+  [Check out our documentation](https://docs.newrelic.com/docs/alerts-applied-intelligence/notifications/notification-integrations/#jira) to set up a JIRA notification channel and provide fast and consistent ways for the right personnel to be notified about incidents
+
+logo: logo.png
```
zstix, line 10:
```suggestion
icon: icon.png
```

data-sources/atlassian-jira/config.yml
zstix: After a discussion about this, we're going to update these fields to match the API more closely (for errors).

data-sources/atlassian-jira/config.yml
```diff
@@ -0,0 +1,31 @@
...
+
+install:
+  primary:
+    nerdlet: nr1-install-newrelic.quickstart-installation-plan
+    nerdletState:
+      quickstart_id: a99b22a5-5666-4096-b239-8f315700fde0
+  requiresAccount: true
+  secondary:
```
aswanson-nr, line 18: The API calls this field `fallback` instead of `secondary`, since we're converting the others to match up, let's change this as well.
```suggestion
  fallback:
```

data-sources/atlassian-jira/config.yml
nr-kkenney: updated

data-sources/atlassian-jira/config.yml
nr-kkenney: updated

data-sources/atlassian-jira/config.yml
nr-kkenney: updated

josephgregoryii (approved):

PR #1041, opened 2022-04-22
chore: update lacework logo

jcountsNR:
updated lacework quickstart logo


quickstarts/lacework/config.yml
```diff
@@ -13,7 +13,7 @@ description: >-
 
   If you encounter any issues, have feedback or would like more details on how to get started, please head over to support.lacework.com or send an email to adam.larson@lacework.net!
 
-logo: logo.svg
+logo: logo.jpg
```
aswanson-nr, line 16: This is the cause of the conflict. We recently changed the name of this field over to `icon` to match our internal schema. Converting it over should resolve the issue.
```suggestion
icon: logo.jpg
```

aswanson-nr (changes requested):
Hey @jcountsNR! Looks like there's a conflict in the config file. My suggested change should deal with it.

josephgregoryii (approved):

aswanson-nr (approved):

PR #1044, opened 2022-04-25
Wayscript Data Source

josephgregoryii:


data-sources/wayscript/config.yml
```diff
@@ -0,0 +1,23 @@
...
+displayName: WayScript
+description: |
+  WayScript is an Internal Developer Platform (IDP) that enables scalable, single click development environments. Connect your environments to New Relic in minutes.
+icon: logo.png
+install:
+  primary:
+    nerdlet:
+      nerdletId: external-install.install
+      nerdletState:
+        quickstart_id: todo-wait-for-quickstart-id
```
aswanson-nr, line 11: Commenting so we don't forget to update this!

PR #1052, opened 2022-04-25
Add partner data sources

aswanson-nr:
# Summary

Adds data sources for many of our partners.


data-sources/cloudflare/config.yml
```diff
@@ -0,0 +1,35 @@
...
+  Cloudflare is on a mission to help build a better Internet. The Cloudflare quickstart will enable you to monitor and analyze web traffic metrics on a dashboard, integrating with New Relic’s database to provide an at-a-glance overview of the most important logs and metrics from your websites and applications.
+
+  For more information check out our website at [cloudflare.com](https://www.cloudflare.com/)
+icon: logo.png
+install:
+  primary:
+    nerdlet:
+      nerdletId: nr1-install-newrelic.quickstart-installation-plan
+      nerdletState:
+        quickstart_id: fc2bb0ac-6622-43c6-8c1f-6a4c26ab5434
```
jerelmiller, line 19: Can we double check this is the right nerdlet state value? I thought this was camelCase, but I could be wrong.

aswanson-nr: @jerelmiller You are correct! It looks like camelCase is the way to go, I'll get these updated.

aswanson-nr: All data sources should be updated to use camelCase

jerelmiller (approved):
This looks great! So stoked to see this in the product soon!

PR #1072, opened 2022-05-05
Additional Datasources

josephgregoryii:
## Summary
Additional datasources added to repo:
* Databricks
* BizTalks360
* ~CodeStream~
* ArgoCD
* Kentik
  - Network Data Ingest and Cardinality
  - Network Flow Devices
  - Network KTranslate Container Health
  - Network Routers and Switches
  - Network Syslog


data-sources/argocd/config.yml
```diff
@@ -0,0 +1,31 @@
+id: argocd
+
+name: ArgoCD
```
zstix, line 3: I think this should be:
```suggestion
displayName: ArgoCD 
```

data-sources/biztalks360/config.yml
```diff
@@ -0,0 +1,24 @@
+id: third-party-biztalk360
+name: BizTalk360
```
zstix, line 2:
```suggestion
displayName: BizTalk360
```

zstix (dismissed):
It looks like the build is failing due to a couple data-sources missing `displayName`.

data-sources/argocd/config.yml
josephgregoryii: Whoops! fixed :)

data-sources/biztalks360/config.yml
josephgregoryii: fixed!

data-sources/codestream/config.yml
```diff
@@ -0,0 +1,17 @@
+id: codestream
+displayName: CodeStream 
+description: |
+  New Relic CodeStream supercharges development workflows by putting collaboration tools in your IDE. It supports pull requests from GitHub, BitBucket and GitLab, issue management from Jira, Trello, Asana and 9 others, observability from New Relic One and Pixie, and provides code discussion that ties it all together, integrated with Slack, MS Teams, email, and in-editor notifications.
+icon: logo.svg
+install:
+  primary:
+    nerdlet:
+      nerdletId: nr1-install-newrelic.quickstart-installation-plan
```
aswanson-nr, line 9: After the discussion yesterday about this not working for quickstarts without install plans, I think we need to sync with Devex on this one to ensure it can be installed properly.

aswanson-nr (commented):
Looks good to me! Just one thing we should check on with the codestream data source before merging.

data-sources/codestream/config.yml
zstix: I think the ID is this, but we should confirm with DevEx: `codestream-install.home`

jerelmiller: `codestream-install.home` is correct! Use that one.

As for when we want to use the quickstart install, the idea behind that was a workaround for the fact that we don't (yet) have auto install dashboards/alerts for data sources. That will be coming, but until then, linking to the quickstart install flow gives us the ability to give our users dashboards/alerts in some form. 

We found out though that blindly doing this can cause problems. Quickstarts without install plans result in a `null` `installer`, but the quickstart install flow expects it to be there. An `installer` is only non-null when there is at least one install plan attached to it.

While you _could_ create an install plan for those, we found these were just links to documentation (no dashboards or alerts), so the quickstart install flow felt a bit weird. It makes you select an account, but we don't do anything with it. For quickstarts that don't have dashboards/alerts, let's make their data sources primary install point to the documentation link directly. You can see an example of that here: https://github.com/newrelic/newrelic-quickstarts/pull/1074

data-sources/codestream/config.yml
```diff
@@ -0,0 +1,17 @@
+id: codestream
```
jerelmiller, line 1: We may or may not want to make this a core data source. I'm working to find out if this should be moved. If so, I'll let you know. We'll need to take this out if that is the case.

jerelmiller: @josephgregoryii we are going to make this a core data source. Can you remove this data source from this PR?

data-sources/codestream/config.yml
josephgregoryii: Resolving this conversation as CodeStream is now a core data-source

data-sources/codestream/config.yml
josephgregoryii: Codestream is now removed!

aswanson-nr (approved):

PR #1080, opened 2022-05-10
fix: [redis-prometheus] Moved Redis dashboard to New Folder

shawnj:
# Summary

The current `Redis (Prometheus)` Quickstart in the I/O catalog doesn't have a dashboard attached.  But the dashboard does exist under `dashboards`.  Assuming the reason for this is that the dashboard should be located under a separate folder under the main `dashboards` folder.

Updated the `redis-prometheus` quickstart to move the `redis-prometheus.json` under a new folder to match the other quickstarts in this directory.


aswanson-nr (approved):
Hey @shawnj, great catch on this!! It must have slipped through when we were migrating the dashboard directories. Thank you!

PR #1093, opened 2022-05-12
NR-2229 Local server for quickstart files

mickeyryan42:
# Summary

Validates the directory path, and statically serves all the files from the quickstart. The response is an array of all file paths.
https://issues.newrelic.com/browse/NR-2229


utils/preview.ts
```diff
@@ -0,0 +1,36 @@
...
+// App
+const app = express();
+const port: number = 3000;
+
+const QUICKSTART_PATH: string = process.argv[2];
+
+/*
+  Validates that the directory path provided exists
+*/
+fs.access(`../quickstarts/${QUICKSTART_PATH}`, (error: any) => {
```
mickeyryan42, line 20: Is there a built in type for error here? I couldn't find the best way to handle it through search.

utils/preview.ts
```diff
@@ -0,0 +1,37 @@
...
+
+app.get('/', (req: Request, res: Response) => {
+  let response = glob.sync(path.join(path.resolve(__dirname, '..'), `/quickstarts/${QUICKSTART_PATH}/**/*.*`))
+                     .map(filePath => filePath.slice(filePath.indexOf('/quickstarts')));
+                     
+  res.send(response);
+});
+
+app.listen(port, () => {
+  console.log("\x1b[32m", `Use this to get a list of your files! http://localhost:3000`);
```
josephgregoryii, line 36:
```suggestion
  console.log("\x1b[32m", `Use this to get a list of your files! http://newrelic.com/instant-observability/preview?local=true`);
```
I think it might be a good idea to have a link to I/O here so when the contributor starts their server, they have a link to click on that will send them to their preview, since it is going to be the browser that fetches from `localhost:3000`.

josephgregoryii (changes requested):

utils/preview.ts
mickeyryan42: Like we discussed on call Friday, updated the output to include the localhost port for the file output as well as the preview link to local I/O.  That will just need to be updated to the hosted I/O env when the feature is complete.

aswanson-nr (approved):

PR #1095, opened 2022-05-16
Add CORS support

mickeyryan42:
# Summary

Adding CORS support for localhost:8000

### DIFF
--- a/utils/preview.ts
+++ b/utils/preview.ts
@@ -6,6 +6,7 @@ import { Request, Response } from 'express';
 // Modules
 import * as path from 'path';
 import * as glob from 'glob';
+import * as cors from 'cors';
 
 // App
 const app = express();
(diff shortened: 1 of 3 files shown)


utils/preview.ts
```diff
@@ -49,14 +50,17 @@ const main = async () => {
 
   await validatePath(QUICKSTART_DIRECTORY);
 
+  app.use(cors({
+    origin: 'http://localhost:8000'
```
aswanson-nr, line 54: We'll probably just want this to use the `port` variable as well.
```suggestion
    origin: `http://localhost:${port}`
```

aswanson-nr: Scratch that, I'm getting the two sides turned around.
But we probably don't want to lock this down to just one URL, that way if it changes we don't need to go back and update this.

aswanson-nr (approved):

utils/preview.ts
mickeyryan42: Got it, updated to just use a wildcard. Thanks!

PR #1103, opened 2022-05-19
NR-2229 Add unit tests

mickeyryan42:
# Summary

Adding unit tests for server preview.  Abstracted the functions to a helper file because jest didn't seem to play nice with exports from the main server file.


utils/preview/previewHelpers.ts
```diff
@@ -0,0 +1,26 @@
...
+      console.error(`Could not find a config.yml or config.yaml at ${fullPath}.`);
+      process.exit(1);
+    }
+  };
+
+/**
+ * Function to remove paths prior to and including /quickstarts
+ * @param {string[]} paths - Array of paths to remove prefixes
+*/
+export const removePathPrefixes = (paths: string[]) => {
```
aswanson-nr, line 24: I think I had a bit of a brain fart on this and forgot which repository we were working in, we actually already have a similar function over in `helpers.ts` https://github.com/newrelic/newrelic-quickstarts/blob/main/utils/helpers.ts#L74-L85

utils/preview.ts
```diff
@@ -48,17 +51,21 @@ const main = async () => {
...
+  app.use(`/quickstarts/${QUICKSTART_DIRECTORY}`, express.static(path.join(PARENT_DIRECTORY, `/${QUICKSTART_DIRECTORY}`)));
 
   app.get('/', (req: Request, res: Response) => {
-    let response = glob.sync(path.join(PARENT_DIRECTORY, `/quickstarts/${QUICKSTART_DIRECTORY}/**/*.*`))
+    let response = glob.sync(path.join(PARENT_DIRECTORY, `/${QUICKSTART_DIRECTORY}/**/*.*`))
                       
-    response = removePathPrefixes(response);                  
+    response = response.map(removeRepoPathPrefix);              
 
     res.json(response);
```
aswanson-nr, line 70: We could simplify this a bit! Totally not a required change though!
```js
const response = glob.sync(path.join(PARENT_DIRECTORY, `/${QUICKSTART_DIRECTORY}/**/*.*`));
res.json(response.map(removeRepoPathPrefix));
```

utils/__tests__/preview_helpers.test.js
```diff
@@ -0,0 +1,38 @@
+import { isPathValid } from '../preview';
+
+import * as path from 'path';
+
+describe('Test preview server helper functions', () => {
+    test('Test isPathValid returns true for valid paths', async () => {
```
aswanson-nr, line 6: If `isPathValid` is no longer async, this can be removed
```suggestion
    test('Test isPathValid returns true for valid paths', () => {
```

utils/__tests__/preview_helpers.test.js
```diff
@@ -0,0 +1,38 @@
...
+            'mock-quickstart-2',
+            'mock-quickstart-3',
+            'mock-quickstart-4',
+            'mock-quickstart-5',
+            'mock-quickstart-6',
+            'mock-quickstart-7/nested-quickstart'
+        ];
+       
+        for (let qs of quickstartPaths) {
+          const isValid = await isPathValid(parentDirectory, qs);
```
aswanson-nr, line 19: If `isPathValid` is no longer async, this can also be removed
```suggestion
          const isValid = isPathValid(parentDirectory, qs);
```

utils/__tests__/preview_helpers.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+    test('Test isPathValid calls console.error', () => {
+        // Arrange
+        const mockConsoleError = jest.spyOn(console, 'error')
+                                     .mockImplementation(() => {});
+        const parentDirectory = path.resolve(__dirname, '../mock_files');
+        const quickstartPath = 'not/valid';
+
+        // Act
+        isPathValid(parentDirectory, quickstartPath);
```
aswanson-nr, line 32: Since we do return `false` in this situation, it's probably worth capturing and asserting on that just to be totally sure.

aswanson-nr (commented):
This looks great! I just noticed a few small changes we could make

aswanson-nr (approved):

zstix (approved):

PR #1117, opened 2022-05-25
Script to create Quickstart mutation snapshots

zstix:
## Description

Adds a script that creates a JSON file in a (git ignored) `snapshots/` directory for _each_ quickstart. We can use these snapshots to ensure that we haven't accidentally mutated any of the quickstarts while moving files around.

To run (from the `utils/` directory):

```sh
$ npx ts-node create-quickstart-snapshots.ts
```

## Related Ticket(s)

- NR-14162

### DIFF
--- a/utils/create-quickstarts-snapshot.ts
+++ b/utils/create-quickstarts-snapshot.ts
@@ -1,0 +1,54 @@
+// To run (from the `utils/` directory):
+// npx ts-node create-quickstarts-snapshot.ts
+
+import * as fs from 'fs';
+import * as path from 'path';
+import {
+  FilePathAndContents,
+  findMainQuickstartConfigFiles,
+  readQuickstartFile,
+} from './helpers';
+import { buildMutationVariables } from './create_validate_pr_quickstarts';
+import { QuickstartConfig } from './types/QuickstartConfig';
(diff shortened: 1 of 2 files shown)


utils/create-quickstarts-snapshot.ts
```diff
@@ -0,0 +1,54 @@
...
+const saveQuickstartMutationVariable = (
+  variable: QuickstartMutationVariable
+) => {
+  const { id, quickstartMetadata } = variable;
+  const { slug } = quickstartMetadata;
+
+  const filename = `${slug}-${id}.json`;
+  const filepath = path.join(__dirname, '..', 'snapshots', filename);
+
+  fs.writeFileSync(filepath, JSON.stringify(quickstartMetadata, null, 2));
```
josephgregoryii, line 34: This fails if `snapshots/` isn't already created - i dont think we're planning on _keeping_ this script long term so might not be an issue

josephgregoryii, line 34: This fails if `snapshots/` isn't already created - i dont think we're planning on _keeping_ this script long term so might not be an issue

aswanson-nr: Considering this will not be run very often, I think it's fine if you have to manually create the `snapshots` directory

PR #1118, opened 2022-05-25
Quickstart Uniqueness JS to TS

josephgregoryii:


utils/check_quickstart_uniqueness.ts
```diff
@@ -0,0 +1,71 @@
+'use strict';
```
aswanson-nr, line 1: We can remove this line now that it's been converted to Typescript!
```suggestion
```

aswanson-nr (dismissed):
Looks good to me!

PR #1125, opened 2022-05-31
NR-827 Convert scripts to TypeScript

mickeyryan42:
# Summary

Converting scripts to typescript
https://issues.newrelic.com/browse/NR-827


utils/check_quickstart_uniqueness.ts
```diff
@@ -1,26 +1,28 @@
...
-import type { FilePathAndContents } from './helpers';
-
-import {
+'use strict';
+const {
   readQuickstartFile,
   removeRepoPathPrefix,
   findMainQuickstartConfigFiles,
-} from './helpers';
+} = require('./helpers');
```
josephgregoryii, line 6: I think it might be more consistent if we use imports here instead of require.

josephgregoryii, line 6: I think it might be more consistent if we use imports here instead of require.

josephgregoryii (changes requested):
Looks great! just a few comments/questions

utils/check_quickstart_uniqueness.ts
mickeyryan42: Strange i dont remember doing this, and not sure why i did.  Let me go switch it back

utils/check_quickstart_uniqueness.ts
```diff
@@ -1,26 +1,28 @@
...
+export type IdsAndPaths = {
+  id: string;
+  path: string;
+};
 /**
  * Returns any quickstarts with matching ids
- * @param idsAndPaths - an array of objects containing the path and id of a quickstart
- * @returns - an array of matching values
+ * @param {Object[]} idsAndPaths an array of objects containing the path and id of a quickstart
+ * @returns {Object[]} an array of matching values
```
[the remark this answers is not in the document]
mickeyryan42: I meant to ask whether we wanted to remove it. Will do

utils/check_quickstart_uniqueness.ts
```diff
@@ -1,26 +1,24 @@
-import type { QuickstartConfig } from './types/QuickstartConfig';
-import type { FilePathAndContents } from './helpers';
-
+'use strict';
```
aswanson-nr, line 1: We can remove this now that it's a typescript file, the compilation step handles setting strict mode for us!

utils/generate-uuids.ts
```diff
@@ -2,20 +2,23 @@
...
 const fs = require('fs');
 const { v4: uuidv4 } = require('uuid');
-const { findMainQuickstartConfigFiles, readYamlFile } = require('./helpers');
+import { findMainQuickstartConfigFiles, readYamlFile } from "./helpers";
+
+import { QuickstartConfig } from './types/QuickstartConfig';
 
 /**
  * Checks for and generates UUIDs for quickstarts
  * @param {String[]} paths an array of quickstarts config paths
```
josephgregoryii: Lets go ahead and remove the types in the JSDocs in this PR 👍🏻

utils/check_quickstart_uniqueness.ts
```diff
@@ -67,4 +61,4 @@ if (require.main === module) {
   main();
 }
 
-module.exports = main;
+module.exports = main;
```
aswanson-nr, line 67: Let's update this export now that we have typescript!
We can totally remove this line
```suggestion
```
And update the `const main = () =>` function to be `export default const main = () =>`

aswanson-nr (commented):
Just a few small cleanup comments around exporting!

josephgregoryii (changes requested):
Just one last think and we should be good to go :)

utils/check-missing-logos.ts
```diff
@@ -0,0 +1,37 @@
+const {
+  readQuickstartFile,
+  findMainQuickstartConfigFiles,
+} = require('./helpers');
```
josephgregoryii, line 4: Looks like we just have one more import that needs to be updated 😄

mickeyryan42: Lol thanks! They were really flying under my radar. I got it removed

josephgregoryii (approved):

josephgregoryii:
Great work on this 😄

aswanson-nr (approved):

PR #1126, opened 2022-05-31
Move dashboard script

josephgregoryii:


utils/lib/dashboards.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type DashboardFileAndPath = FilePathAndContents<Dashboard>;
+
+export const getAllDashboardPaths = () => {
+  return findQuickstartComponentConfiguration('dashboards');
+};
+
+// .sjon
+// .png
+// assets/iconm.svg
```
zstix, line 23: Do we need these in here? If so, it might be good to add some more context (and fix some spelling mistakes 😄)

josephgregoryii: Removed and included documentation for functions

utils/migration_scripts/move-dashboards.ts
```diff
@@ -0,0 +1,161 @@
...
+    );
+    const dashboardScreenshots = dashboardScreenshotPaths.map((screenshot) => {
+      return {
+        oldFilePath: screenshot,
+        newFilePath: path.join(newDashboardLocation, path.basename(screenshot)),
+      };
+    });
+    // create new files in this location
+    dashboardScreenshots.forEach(({ oldFilePath, newFilePath }) =>
+      fs.copyFileSync(oldFilePath, newFilePath)
```
aswanson-nr, line 105: Could probably use the `copyFiles` function here, not a required change at all.

aswanson-nr (approved):
Looking good to me!

tabathadelane (approved):

aswanson-nr (approved):

PR #1226, opened 2022-06-07
Update install plan for Fluentd

josephgregoryii:


zstix (approved):

aswanson-nr:
@josephgregoryii Could you also put a small update into the corresponding quickstart so that it gets published as well?

aswanson-nr (approved):

PR #1246, opened 2022-06-10
fix: properly handle untrusted input

aswanson-nr:
# Summary
* Handles untrusted input in the validation and submission workflows by
  evaluating it as a environment variable before using it within a
  script, this prevents command injection
* Also changes the "Add status to pull request" to grab the event
  information as an environment variable instead of using jq parsing

### DIFF
--- a/.github/workflows/validate_data_sources.yml
+++ b/.github/workflows/validate_data_sources.yml
@@ -47,15 +47,17 @@ jobs:
           NEW_RELIC_APP_NAME: ${{ secrets.NEW_RELIC_APP_NAME }}
           NEW_RELIC_HOST: staging-collector.newrelic.com
           NODE_ENV: production
+          PR_NUMBER: ${{ steps.get_pr_number.outputs.pr-number }}
         run: |
-          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ steps.get_pr_number.outputs.pr-number }}/files"
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/$PR_NUMBER/files"
           DRY_RUN=true
           cd utils && yarn create-validate-data-sources $URL $DRY_RUN
 
       - name: Add commit status
         if: always()
+        env:
+          COMMIT_SHA: ${{ github.event.workflow_run.head_commit.id }}
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/validate_data_sources.yml
```diff
@@ -47,15 +47,17 @@ jobs:
...
 
       - name: Add commit status
         if: always()
+        env:
+          COMMIT_SHA: ${{ github.event.workflow_run.head_commit.id }}
         run: |
-          export COMMIT_SHA=$(echo '${{ toJSON(github.event) }}' | jq '.workflow_run.head_commit.id' | tr -d '"')
           curl \
             -X POST \
             --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
```
josephgregoryii, line 63: Does `{{ secrets.GITHUB_TOKEN }}` need to change to use the env variable instead?

.github/workflows/validate_install_plans.yml
```diff
@@ -47,15 +47,17 @@ jobs:
...
 
       - name: Add commit status
         if: always()
+        env:
+          COMMIT_SHA: ${{ github.event.workflow_run.head_commit.id }}
         run: |
-          export COMMIT_SHA=$(echo '${{ toJSON(github.event) }}' | jq '.workflow_run.head_commit.id' | tr -d '"')
           curl \
             -X POST \
             --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
```
josephgregoryii, line 63: here too?

.github/workflows/validate_packs.yml
```diff
@@ -141,7 +142,6 @@ jobs:
       - name: Add commit status
         if: always()
         run: |
-          export COMMIT_SHA=$(echo '${{ toJSON(github.event) }}' | jq '.workflow_run.head_commit.id' | tr -d '"')
           curl \
             -X POST \
             --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
```
josephgregoryii, line 147: ^

.github/workflows/validate_data_sources.yml
aswanson-nr: That value is generated by Github when the job runs, so it is considered trusted input. So we are not worried about it being escaped.

josephgregoryii (approved):
Thanks for clarifying 👍🏻

PR #1255, opened 2022-06-22
Repo Restructure and Refactor

josephgregoryii:


utils/migration_scripts/move-dashboards.ts
```diff
@@ -1,25 +1,24 @@
...
 } from '../lib/dashboards';
 
-import { getQuickstartFromFilename } from '../helpers';
-
-import * as path from 'path';
-import * as fs from 'fs';
-
 import { RmDirOptions } from 'fs';
-import * as glob from 'glob';
+import { getQuickstartFromFilename } from '../helpers';
```
josephgregoryii, line 13: We need to add this into `helpers` or use what is already in `lib/helpers`

utils/lib/Component.ts
```diff
@@ -0,0 +1,51 @@
+import * as path from 'path';
+import * as fs from 'fs';
+import * as yaml from 'js-yaml';
+
+abstract class Component<ConfigType, MutationVariablesType> {
+  public localPath: string; // Local path to the component. Ex: python/flask
```
zstix, line 6: I worry that this variable isn't super clear what it represents. Perhaps we should call this `identifier`? `slug`?

utils/lib/Component.ts
```diff
@@ -0,0 +1,51 @@
...
+  public basePath: string;
+
+  constructor(
+    localPath: string,
+    basePath: string = path.join(__dirname, '../..')
+  ) {
+    this.basePath = basePath;
+    this.localPath = localPath;
+    this.configPath = this.getConfigFilePath();
+    this.config = this.getConfigContent();
```
zstix, line 19: We have custom functions for these two properties, but a getter for `fullPath`. Just something to consider.

utils/lib/Component.ts
```diff
@@ -0,0 +1,51 @@
...
+abstract class Component<ConfigType, MutationVariablesType> {
+  public localPath: string; // Local path to the component. Ex: python/flask
+  public configPath: string; // Absolute path to the config file within the repository
+  public config: ConfigType;
+  public isValid = true;
+  public basePath: string;
+
+  constructor(
+    localPath: string,
+    basePath: string = path.join(__dirname, '../..')
```
zstix, line 14: If we _wanted_ to, we could use a shorthand here: https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd

utils/lib/Component.ts
```diff
@@ -0,0 +1,51 @@
...
+    return path.join(this.basePath, this.configPath);
+  }
+
+  /**
+   * Helper method for reading YAML-based configuration files.
+   */
+  protected _getYamlConfigContent(): ConfigType {
+    if (!this.isValid) {
+      return this.config;
+    }
```
zstix, line 36: This is a tad misleading. If the component isn't valid, then returning `this.config` will return `undefined` here. Perhaps we should update the generic to be `ConfigType | undefined`.

utils/lib/Dashboard.ts
```diff
@@ -0,0 +1,112 @@
...
+
+interface DashboardConfig {
+  name: string;
+  description?: string;
+  pages: any;
+}
+
+class Dashboard extends Component<DashboardConfig, QuickstartDashboardInput> {
+  /**
+   * Returns the file path from the top level of component
```
zstix, line 19:
```suggestion
```

utils/lib/classHelpers.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param basePath - The portion of the path to remove
+ * @returns - The path with the prefix
+ */
+export const removeBasePath = (filePath: string, basePath: string): string => {
+  const shortPath = filePath.split(`${basePath}/`).pop();
+  if (typeof shortPath === 'string') {
+    return shortPath;
+  }
+  return filePath;
+};
```
zstix, line 24: I find this to be a tad odd that this lives outside of our `helpers.ts` file _and_ outside of our base `Component.ts` class. Perhaps we should look into adding it to the base class as a `static` method (or not static, if possible).

utils/lib/classHelpers.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param filePath - The path to change
+ * @param basePath - The portion of the path to remove
+ * @returns - The path with the prefix
+ */
+export const removeBasePath = (filePath: string, basePath: string): string => {
+  const shortPath = filePath.split(`${basePath}/`).pop();
+  if (typeof shortPath === 'string') {
+    return shortPath;
+  }
+  return filePath;
```
zstix, line 23: Another way to handle this would be with a [short-curcuit `||` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#short-circuit_evaluation), since `pop` is always going to return a string or a falsy value:
```ts
  return shortPath || filePath;
```
Another option would be the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator):
```suggestion
return shortPath ?? filePath
```
The second would keep the functionality the same

utils/lib/DataSource.ts
```diff
@@ -0,0 +1,149 @@
...
+        },
+      };
+    }
+
+    if ('nerdlet' in directive) {
+      const { nerdletId, nerdletState, requiresAccount } = directive.nerdlet;
+
+      return {
+        nerdlet: {
+          nerdletId: nerdletId.trim() ?? '',
```
aswanson-nr, line 138: This should throw a `cannot reference trim() of undefined` error if the `nerdletId` is `undefined`. We can use optional chaining to deal with this.
```suggestion
          nerdletId: nerdletId?.trim() ?? '',
```

utils/lib/InstallPlan.ts
```diff
@@ -0,0 +1,173 @@
...
+  ): InstallPlanMutationVariable {
+    const { id, description, name, title, install, fallback, target } =
+      this.config ?? {};
+
+    return {
+      dryRun,
+      id,
+      description,
+      displayName: name,
+      heading: title,
```
aswanson-nr, line 107: Let's trim these to be safe
```suggestion
      displayName: name && name.trim(),
      heading: title && title.trim(),
```

utils/lib/Quickstart.ts
```diff
@@ -0,0 +1,269 @@
...
+  'New Relic': 'NEW_RELIC',
+  Community: 'COMMUNITY',
+  Verified: 'VERIFIED',
+};
+
+type ComponentType =
+  | typeof Alert
+  | typeof Dashboard
+  | typeof DataSource
+  | typeof InstallPlan;
```
aswanson-nr, line 47: Since install plans are not being pull into this class, let's remove it.
```suggestion
  | typeof DataSource;
```

utils/validate_images.ts
```diff
@@ -1,106 +1,68 @@
...
+    const dashboardDirName = path.dirname(dashboard.configPath);
 
-    const dashboardImagePaths = imagePaths.filter(
-      (p: string[]) =>
-        p !== logoPath && p.includes(quickstartName + DASHBOARD_IMAGES_PATH)
-    );
+    const dashboardImagePaths = glob
+      .sync(path.join('..', dashboardDirName))
+      .filter((filePath) => ALLOWED_IMG_EXT.includes(path.extname(filePath)))
+      .filter((filePath) => IGNORE_FILE_EXT.includes(path.extname(filePath)));
```
aswanson-nr, line 39: We might be missing something here, but this appears to filter down to just image files, then filter again down to just JSON files, which should leave us with an empty array. 
If we're correct, then we can update it to:
```suggestion
      .sync(path.join('..', dashboardDirName))
      .filter((filePath) => ALLOWED_IMG_EXT.includes(path.extname(filePath)));
```

utils/validate_images.ts
```diff
@@ -1,106 +1,68 @@
...
+  const imagesDirectories: DirectoryValidation[] = [];
+  dashboards.forEach((dashboard) => {
+    const dashboardDirName = path.dirname(dashboard.configPath);
 
-    const dashboardImagePaths = imagePaths.filter(
-      (p: string[]) =>
-        p !== logoPath && p.includes(quickstartName + DASHBOARD_IMAGES_PATH)
-    );
+    const dashboardImagePaths = glob
+      .sync(path.join('..', dashboardDirName))
```
aswanson-nr, line 37: This path should be `dashboards/dashboard-name` after the join. Which is not a glob pattern that recursively grabs all the files in the directory, it will just return the directory name itself. I think we need to add a `*` pattern so that it pulls in all the required files.
```suggestion
      .sync(path.join('..', dashboardDirName, '**', '*'))
```

utils/create-validate-data-sources.ts
```diff
@@ -267,3 +56,5 @@ const main = async () => {
 if (require.main === module) {
   main();
 }
+
+export default main;
```
aswanson-nr, line 60: Are we sure we want to export a default from this file? If it's just being run as a script, there is no need to export the `main` function

aswanson-nr (commented):
Overall looks pretty much good to go! Just saw one or two spots that absolutely need to be updated before merging

utils/validate_images.ts
josephgregoryii: This is resolved - we now filter out any JSON files before being sent to the validation functions

utils/validate_images.ts
josephgregoryii: Resolved 👍🏻

utils/lib/Component.ts
aswanson-nr: Updated to use `identifier`

utils/lib/classHelpers.ts
aswanson-nr: Moved these to static methods

utils/lib/classHelpers.ts
aswanson-nr: Done!

aswanson-nr (approved):

PR #1266, opened 2022-06-29
Local preview server updated to support new structure

mickeyryan42:
# Summary
https://issues.newrelic.com/browse/NR-26664
Now serving all files.  GET request will now give paths to dashboards and alerts to appropriate new directories.  Removed the tests for validating paths because this is tested in the class.


aswanson-nr (approved):

josephgregoryii (approved):

aswanson-nr:
I think you'll want to target the `content-updates` branch with this, that's probably why we're seeing conflicts

mickeyryan42:
> I think you'll want to set `content-updates` as the base branch with this, that's probably why we're seeing conflicts

Ahh I didn't know that existed, my mistake.  I pointed it to the new base, that seems to have resolved the open conflicts so I'll merge this

PR #1275, opened 2022-07-05
NR-30287: --frozen-lockfile in all workflows

mickeyryan42:
https://issues.newrelic.com/browse/NR-30287


aswanson-nr (commented):
Looks good, just noticed one missed spot in the [release workflow](https://github.com/newrelic/newrelic-quickstarts/blob/main/.github/workflows/release.yml#L30)

mickeyryan42:
> Looks good, just noticed one missed spot in the [release workflow](https://github.com/newrelic/newrelic-quickstarts/blob/main/.github/workflows/release.yml#L30)

Good catch, I missed saving that file.  Added it in

aswanson-nr (approved):

PR #1295, opened 2022-07-26
Query categoryTerms from Graphql

josephgregoryii:
### Summary
Earlier this sprint, we had a [bug](https://issues.newrelic.com/browse/NR-3896) in the internal I/O where dashboards and alerts were not rendering under associated tabs. Specifically, the `partner` category.

We had originally used a [hard coded file](https://github.com/newrelic/newrelic-quickstarts/blob/0421c87883c3f35c6ac3449971fc412de1d05e69/utils/instant-observability-categories.ts) to match the category terms with the associated tabs for quickstarts. Before a fix, the categories file was outdated and needed an update, along with mutations submitted to graphql to update those quickstarts.

This PR allows us to fetch the category terms during submission to avoid losing category terms in the future, while also removing the category files


utils/lib/Quickstart.ts
```diff
@@ -125,9 +125,11 @@ class Quickstart {
 
   /**
    * Get mutation variables from quickstart config
-   * @returns - mutation varaibles for quickstart
+   * @returns - Promised mutation variables for quickstart
    */
-  getMutationVariables(dryRun: boolean): QuickstartMutationVariable {
+  async getMutationVariables(
```
aswanson-nr, line 130: We'll want to make sure that is function is now being awaited wherever it's used throughout the scripts

utils/lib/nr-graphql-helpers.ts
```diff
@@ -124,10 +124,22 @@ export const translateMutationErrors = (
   }
 };
 
+type CategoryTermsNRGraphqlResults = {
+  actor: {
+    nr1Catalog: {
+      categories: {
+        displayName: string;
+        slug: string;
```
aswanson-nr, line 132: Are these fields even needed? On lines 165-167 we are just flat mapping down to the terms.

aswanson-nr (commented):
I found a few things worth checking on, but overall looks fantastic!

utils/lib/Quickstart.ts
josephgregoryii: Weird, i must've restored at some point and lost the changes i made. I'll go back and fix this!

utils/lib/nr-graphql-helpers.ts
josephgregoryii: We do not - should i update the query to reflect this as well?

utils/lib/Quickstart.ts
josephgregoryii: Only needed the fix inside `create-quickstarts-snapshot.ts` 👍🏻 . `create_validate_pr_quickstart` has the function calls mapped inside a `Promise.all`

utils/lib/nr-graphql-helpers.ts
josephgregoryii: updated on the typescript side of things 👍🏻

aswanson-nr: Might as well remove those from the query as well, no need to request more than we need!

aswanson-nr (approved):

zstix (approved):

PR #1305, opened 2022-08-01
Improved error handling for quickstart components

josephgregoryii:
### Summary
This PR updates some error handling across classes and the quickstart validation workflow.

Improvements include:
- Improved console logging for validating quickstart and associated components
- Validation workflow keeps track of any errors thrown from methods to fail workflow checks
-  `validate()` function is now called in workflow to validate associated components and quickstart


utils/check_dashboard_name_uniqueness.ts
```diff
@@ -22,21 +22,40 @@ const findMatchingDashboardNames = (
...
-    console.error(`ERROR: Found matching quickstart dashboard names`);
-    console.error(`Punctuation and white space are removed before comparison`);
-    nameMatches.forEach((m) =>
-      console.error(`${m.identifier} in ${removeRepoPathPrefix(m.configPath)}`)
-    );
-    console.error(
-      `Please update your quickstart dashboard's name to be unique\n`
-    );
+  const catchErrors = []
+  try {
```
aswanson-nr, line 26: Is this try/catch block even needed here anymore? It doesn't look like any of these methods throw an error.

utils/create_validate_pr_quickstarts.ts
```diff
@@ -53,17 +53,41 @@ const main = async () => {
...
+  const invalidQuickstarts = quickstarts
+    .map((qs) => {
+      qs.validate();
+      if (qs.isValid) {
+        return qs;
+      } else {
+        return undefined;
+      }
+    })
+    .filter(Boolean);
```
aswanson-nr, line 65: This could be simplified down to 
```js
const invalidQuickstarts = quickstarts
  .map((qs) => qs.validate())
  .filter((qs) => !qs.isValid);
```
However, that does make us loop through the array twice, so totally up to you which direction you'd like to go.

utils/lib/Alert.ts
```diff
@@ -51,16 +57,25 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
       this.isValid = false;
 
       return this.config;
     }
   }
 
   getMutationVariables() {
-    if (!this.isValid) {
-      return [];
+    const isPathGiven = this.configPath === ''
```
aswanson-nr, line 68: Shouldn't the Alert be considered invalid if the config path is the empty string?

aswanson-nr (commented):
Overall looks good! I had a few comments, but nothing major. One thing I would like to see is some test cases that cover the original bug case, i.e. a quickstart config file having invalid references to components

utils/create_validate_pr_quickstarts.ts
josephgregoryii: I tried this approach however `.validate()` doesn't return anything, so `.filter()` receives nothing.

I went with the first approach, just a little cleaner instead

utils/lib/Alert.ts
```diff
@@ -22,6 +22,12 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
 
     if (!Array.isArray(filePaths) || filePaths.length !== 1) {
       this.isValid = false;
+      const errorMessage =
+        filePaths.length > 1
+          ? `Alert at ${this.identifier} contains multiple configuration files.\n`
+          : `Alert at ${this.identifier} does not exist. Please double check this location.\n`;
```
aswanson-nr, line 28: I think we do expect there to be multiple alert conditions per policy, so this logic will need a little update

utils/lib/Quickstart.ts
```diff
@@ -181,7 +186,7 @@ class Quickstart {
       id: id ? id : MOCK_UUID,
       dryRun,
       quickstartMetadata,
-    };
+    } as QuickstartMutationVariable;
```
aswanson-nr, line 189: Typescript should determine this automatically
```suggestion
    };
```

aswanson-nr (commented):
A few small comments, but looks good!

utils/lib/Alert.ts
josephgregoryii: fixed

utils/lib/Quickstart.ts
josephgregoryii: done

aswanson-nr (approved):

mickeyryan42 (approved):

PR #1322, opened 2022-08-16
Fix preview link issue, fix comment formatting

mickeyryan42:
https://issues.newrelic.com/browse/NR-37928


utils/create-preview-links.ts
```diff
@@ -23,30 +23,27 @@ export const getQuickstartsFromPRFiles = async (
...
+  
+  const quickstartPaths = nonTestFiles
+      .map(prop('filename'))
+      .filter(filePath => QUICKSTART_CONFIG_REGEXP.test(filePath))
+      .map(filePath => path.dirname(filePath.split('quickstarts/').pop()!))
+
+  /*
+    We can rely on a quickstart existing, and use the quickstart class with 
+    getRelatedQuickstarts to get all quickstarts related to the component.  If
+    the component itself exists we know a quickstart related to it also exists.
```
aswanson-nr, line 37:
```suggestion
    We can rely on a quickstart existing on the `main` branch because if it isn't, then it was created in this 
    pull request and will be caught by the logic above. We can then use the Quickstart class with 
    getRelatedQuickstarts to get all quickstarts related to the component.  If
    the component itself exists we know a quickstart related to it also exists.
```

aswanson-nr (commented):
Looks great to me! Just one suggestion on the comment to make it clear why this works for us.

zstix (approved):

aswanson-nr (approved):

zstix (approved):

PR #1341, opened 2022-08-26
Add keywords for Node.js

aswanson-nr:
# Summary

Adds some more variations of `node.js` to the Node.js quickstart keyword list to improve search results.


tabathadelane (approved):
good call 👍

PR #1342, opened 2022-08-29
vercel datasource

jcountsNR:
Adding a vercel datasource and updating the logo.


data-sources/vercel/config.yml
```diff
@@ -0,0 +1,34 @@
...
+  ## About this quickstart
+
+  This quickstart is the fastest way to jump straight into monitoring your end-to-end development logs.
+icon: logo.svg
+install:
+  primary:
+    nerdlet:
+      nerdletId: nr1-install-newrelic.quickstart-installation-plan
+      nerdletState:
+        quickstartId: fc2bb0ac-6622-43c6-8c1f-6a4c26ab5434
```
aswanson-nr, line 19: This is the data source related to the "Vercel Logs" quickstart, correct?
If so, this quickstart id should be `c18372e0-b849-408f-b2b1-3ea7096ef5a8`
```suggestion
        quickstartId: c18372e0-b849-408f-b2b1-3ea7096ef5a8
```

aswanson-nr (commented):
Hey @jcountsNR, looks good I just think your might have an incorrect quickstart id in the data source!

jcountsNR:
@aswanson-nr ah good catch! I thought I had pasted over that but looks like it didn't copy properly.

aswanson-nr (approved):

PR #1349, opened 2022-09-06
Dashboard PR helper script

aswanson-nr:
# Summary

Adds a new Github workflow that will alert a contributor via PR comment of invalid fields in dashboard JSON files. 
Also includes unit testing.


aswanson-nr:
Not done with this yet!

utils/dashboard-helper.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+  if (permissionsFieldRegex.test(line)) {
+    warningsFound.push(`\"permissions\" field should not be used`);
+  }
+  if (accountIdRegEx.test(line)) {
+    warningsFound.push(`\"accountId\" must be zero`);
+  }
+
+  return warningsFound;
+};
```
josephgregoryii, line 30: [nit] - Not a blocker at all, but I'm wondering if there is a way we can simplify this code to avoid so many `if` statements. In the chance we need to add more regEx in the future, having to add in the regex in 2 different places seems odd to me.

I put a suggestion below, but feel free to disregard it

```suggestion
const regexAndWarning = [
  [ /guid[`'"\) ]/, `\"guid\" should not be used` ],
  [ /entityGuid/, `\"entityGuid\" should not be used` ],
  [ /\"linkedEntityGuids\": (?:(?!null))/, `\"entityGuid\" should not be used`],
  [ /\"permissions\": /, `\"permissions\" field should not be used` ],
  [ /\"accountId\": (?:(?!0))/, `\"accountId\" must be zero` ],
]

export const checkLine = (line: string) => {
  const warningsFound = [];

  for const [regex, warning] of regexAndWarning {
    if (regex.test(line)) {
      warningsFound.push(warning)
    }
  }

  return warningsFound;
};
```

josephgregoryii (dismissed):
Looks solid! - small nit but not a blocker

utils/dashboard-helper.ts
tabathadelane: @josephgregoryii I like it! I made these changes. Tests still pass, and it seems to be working as expected locally. I think it's ready for a re-review.

josephgregoryii (approved):
AWESOME

tabathadelane (commented):
I LOVE IT! GREAT WORK

mickeyryan42 (approved):

PR #1382, opened 2022-09-26
chore: update dashboard image for apprunner

jcountsNR:
update image for aws app-runner


jcountsNR:
@aswanson-nr , Hi, I was asked to see if we can go ahead and push this through, I thought you might be the one to review since you reviewed the main PR. It's only an image change, looks like we had the wrong image for the dashboard.

cc @4demos

aswanson-nr (approved):

PR #1383, opened 2022-09-26
Filter out removed files during PR validation

aswanson-nr:
# Summary

Removed files were not being filtered out during validation, which can cause errors when the scripts attempt to read the file on disc.
This updates the logic for quickstart, install plan, and data-source validation.

### DIFF
--- a/utils/__tests__/create-validate-pr-quickstarts.test.js
+++ b/utils/__tests__/create-validate-pr-quickstarts.test.js
@@ -138,4 +138,14 @@ describe('create-validate-pr-quickstarts', () => {
     const hasErrored = await createValidateQuickstarts('url', 'token');
     expect(hasErrored).toBe(false);
   });
+
+  test('does not process removed file', async () => {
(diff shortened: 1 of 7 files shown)


tabathadelane:
Looks great but curious if [dashboard-helpers](https://github.com/newrelic/newrelic-quickstarts/blob/alec/ignore-removed-files/utils/dashboard-helper.ts#L62) (and others if they exists) should be updated to use this new github-api-helper function also?

josephgregoryii (dismissed):
Good catch!

zstix (approved):

aswanson-nr:
> Looks great but curious if [dashboard-helpers](https://github.com/newrelic/newrelic-quickstarts/blob/alec/ignore-removed-files/utils/dashboard-helper.ts#L62) (and others if they exists) should be updated to use this new github-api-helper function also for consistency?

Good call! I found 1 other place as well and updated the PR!

tabathadelane (approved):

josephgregoryii (approved):

PR #1470, opened 2022-10-26
fix: Small change to query that returns logs by app

JimHagan:
# Summary

Made a small refinement to three queries that return logs by app.  The old model may not return results if the customer has DT turned off.  The new query will return results for apps regardless of whether they have DT on or off.


dashboards/log-analysis/log-analysis.json
```diff
@@ -420,7 +420,7 @@
             "nrqlQueries": [
               {
                 "accountId": 0,
-                "query": "SELECT count(*) FROM Log WHERE trace.id IS NOT NULL since 1 hour ago facet entity.name limit 100"
+                "query": "SELECT count(*) FROM Log WHERE entity.guid IS NOT NULL since 1 hour ago facet entity.name limit max"
```
josephgregoryii, line 423: @JimHagan - Since we're filtering on `entity.guid`, this may not work for another user's account if they attempt to install this dashboard. Its a small caveat, however we would need to have a more specific variable to replace `entity.guid`.

JimHagan: @josephgregoryii 

We are only checking to see if the entity.guid is NULL not checking the absolute value of the GUID.  We know that if there is an entity.guid we can facet by entity name.

aswanson-nr: Sounds good to me!

aswanson-nr (approved):
Looks good to me! We'll get this merged on Monday!

PR #1471, opened 2022-10-27
Infra Template Migration

lroyce:
# Summary

Updated several dashboard and quickstart templates with infra dashboard templates. These include:

1. AWS-Elastic Beanstalk
2. AWS-Appsync
3. AWS-Glue
4. Amazon-Athena
5. Amazon-FSX

AWS Elastic Beanstalk (environments) was combined into a tabbed dashboard experience with the Elastic Beanstalk (instances).
AWS-FSX and AWS-Athena naming was updated to Amazon FSx and Amazon Athena per AWS guides and docs. 

Install plans were updated for all 4 template configs as they have cloudwatch metrics available. 

 Amazon Athena template 'athena.json' was merged with the Infra Dashboard template based on Metrics. There were extraneous properties in the JSON from the athena.json which were removed.

## Pre merge checklist

- [x] Are all documentation links publically accessible?
- [x] Did you check your descriptive content for [voice and tone](https://docs.newrelic.com/docs/style-guide/writing-strategies/voice-strategies-docs-sound-new-relic/)? 
- [x] Did you check your descriptive content for spelling and grammar errors?

### Dashboards 

- [x] Does the PR contain a screenshot for each of your dashboards?
- [x] Do your screenshots show data?
- [x] Has the [sanitization script](https://github.com/newrelic/newrelic-quickstarts/blob/main/CONTRIBUTING.md#dashboards) been run on each dashboard?

### Alerts


lroyce:
amazon-athena.json and athena.json are both for AWS Athena, but the one I've added is Dimensional Metrics based, the athena.json which existed in IO is Sample Events based. I'm not sure if we want to remove one or leave both?

quickstarts/aws/amazon-athena/config.yml
```diff
@@ -37,4 +37,9 @@ keywords:
   - amazon web services
   - database
 installPlans:
-  - aws-cloudwatch-metric-streams
\ No newline at end of file
+  - aws-cloudwatch-metric-streams
+
+dashboards:
+  - athena
```
josephgregoryii, line 43: I think this may be an accidental duplicate!

```suggestion
```

josephgregoryii (commented):
This looks great! One comment and then i can merge this in

quickstarts/aws/amazon-athena/config.yml
lroyce: This is actually related to the question on whether we should include the existing athena/athena.json which is Sample Metric based vs the amazon-athena/amazon-athena.json that is migrated from infra templates and is dimensional metrics based. I'm not sure if we want to keep both? The sample metric json has more charts, the dimensional metrics only has 2. I'm not sure if it's possible to leave both so users have the choice?

aswanson-nr: Hey @lroyce, is it possible to merge the two dashboards? That way we get the best of both worlds!

lroyce: @aswanson-nr that's definitely doable: would we put all together in a single dashboard? Or would we want a tabbed dashboard, one for metrics, one for samples?

aswanson-nr: My understanding is that there is no difference between samples and metrics to the end user when they're looking at the dashboard. If that's correct, then let's just toss them all on one tab. If not, then separate tabs sounds great!

lroyce: @aswanson-nr and @josephgregoryii `amazon-athena.json` was merged with `athena.json` and now has both widget queries.

josephgregoryii (approved):

PR #1546, opened 2022-11-30
Use data source ids in quickstart definition

josephgregoryii:
## Summary
- Update `CONTRIBUTING` doc to represent `dataSourceIds`
- `DataSource` class is able to be passed in an ID instead of the local path
- `Quickstart` class accepts `dataSourceIds` as a config field and adds it to mutation variables
- Updated tests to specify new changes to `DataSource` class
- New validation script to check against dataSourceIds


CONTRIBUTING.md
```diff
@@ -175,7 +175,7 @@ icon.png
 | dashboards    | no        |                                                                   | A list of dashboards to include in this quickstart                                                         |
 | alertPolicies | no        |                                                                   | A list of alert policies to include in this quickstart                                                     |
 | installPlans  | no        |                                                                   | A list of install plans, at least one is required to allow this quickstart to be installed                 |
-| dataSources   | no        |                                                                   | A list of data sources                                                                                     |
+| dataSourceIds | no        |                                                                   | A list of data sources                                                                                     |
```
aswanson-nr, line 178:
```suggestion
| dataSourceIds | no        |                                                                   | A list of data sources Ids                                                                                     |
```

aswanson-nr (commented):
Saw a small issue with the docs, but otherwise looks good to go!

utils/package.json
```diff
@@ -6,6 +6,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "validate-quickstart-install-plans": "ts-node validate-quickstart-install-plans.ts",
+    "validate-quickstart-data-sources": "ts-node validate-quickstart-data-sources.ts",
```
aswanson-nr, line 9: Looks like the file is actually called `validate-quickstart-data-source.ts` (with out `s`), let's update the filename to be plural.

utils/validate-quickstart-data-source.ts
```diff
@@ -0,0 +1,68 @@
...
+  githubFiles: GithubAPIPullRequestFile[]
+) => {
+  const quickstartsWithInvalidDataSources = filterQuickstartConfigFiles(
+    githubFiles
+  )
+    .filter(isNotRemoved)
+    .map(({ filename }) => new Quickstart(filename))
+    .map((quickstart)=> {
+      const invalidIds = quickstart.config.dataSourceIds?.filter((dataSourceId)=> {
+        return !new DataSource(dataSourceId).isValid;
```
aswanson-nr, line 23: If I attempt to use a core data source on a quickstart, this will fail. Could we possibly pull in the list of data sources to check against from Nerdgraph?
The only other option I see is letting this fail for core data sources and getting the success from the validation against Nerdgraph. However, I can't think of a way to say that one or the other github actions must succeed to merge.

utils/validate-quickstart-data-source.ts
```diff
@@ -0,0 +1,68 @@
+import {
```
aswanson-nr, line 1: I could see some unit tests being helpful for this file. Especially if it's going to pull from nerdgraph (see other comment!)

aswanson-nr (changes requested):

utils/lib/nr-graphql-helpers.ts
```diff
@@ -175,6 +175,32 @@ export const getCategoryTermsFromKeywords = async (
...
+        }[];
+      };
+    };
+  };
+};
+
+export const getCoreDataSourceIds = async (): Promise<string[]> => {
+  const { data } = await fetchNRGraphqlResults<{}, CoreDataSourceSearchResults>(
+    { queryString: CORE_DATA_SOURCES_QUERY, variables: {} }
+  );
```
aswanson-nr, line 193: This function can possibly return errors, so we should handle those errors if we get them. 
I think you have a few options here, we could either throw the error and reject the promise, or just log out any errors and return an empty array.

utils/lib/nr-graphql-helpers.ts
```diff
@@ -175,6 +175,32 @@ export const getCategoryTermsFromKeywords = async (
...
+      search: {
+        results: {
+          id: string;
+        }[];
+      };
+    };
+  };
+};
+
+export const getCoreDataSourceIds = async (): Promise<string[]> => {
```
aswanson-nr, line 190: I think this name is a little misleading, since the query will actually return all data sources. Perhaps something like 
```suggestion
export const getPublishedDataSourceIds = async (): Promise<string[]> => {
```

utils/validate-quickstart-data-sources.ts
```diff
@@ -0,0 +1,80 @@
...
+import { getCoreDataSourceIds } from './lib/nr-graphql-helpers';
+
+import Quickstart from './lib/Quickstart';
+import DataSource from './lib/DataSource';
+import { passedProcessArguments } from './lib/helpers';
+
+export const validateDataSourceIds = async (
+  githubFiles: GithubAPIPullRequestFile[]
+) => {
+  const coreDataSourceIds = await getCoreDataSourceIds();
```
aswanson-nr, line 18: We'll need to handle the error case for this function. Depending on what direction you choose, we could either add a try catch or just exit if no data source ids are returned.

aswanson-nr (changes requested):
Great additions! I think there's a little more we can do around error handling here, since we are calling out to nerdgraph. If that was to fail, I'm thinking it would be really difficult to track down.

aswanson-nr (dismissed):

utils/lib/nr-graphql-helpers.ts
```diff
@@ -175,6 +175,39 @@ export const getCategoryTermsFromKeywords = async (
   return categoryKeywords.length > 0 ? categoryKeywords : undefined;
 };
 
+type CoreDataSourceSearchResults = {
```
caylahamann, line 178: @MichelLosier and I were talking about this while working on dashboards but it might be nice in the future if we could use codegen in order to generate these types from our queries (similar to how we do it in the marketplace?)

caylahamann (approved):

utils/__tests__/validate-quickstart-datasource.test.js
```diff
@@ -0,0 +1,220 @@
...
+    githubHelpers.filterQuickstartConfigFiles.mockReturnValueOnce(files);
+    
+    nrGraphQlHelpers
+      .getPublishedDataSourceIds
+      .mockResolvedValueOnce({errors: [new Error('test error')]});
+
+
+    nrGraphQlHelpers
+      .getPublishedDataSourceIds
+      .mockResolvedValueOnce({erros: [new Error('fake error')]})
```
MichelLosier, line 118:
```suggestion
      .mockResolvedValueOnce({errors: [new Error('fake error')]})
```

MichelLosier (changes requested):
Just a small typo and should be good to go!

MichelLosier (approved):

PR #1548, opened 2022-12-01
feat(prom-agent): duplicated data alert

paologallinaharbur:
# Summary

Due to a new version of the integrations filters, a target that was already scaped by a different job is scraped again. 
In order to prevent that and be notified in case of duplicated data, you can create an alert based on the following query:
```
FROM Metric select uniqueCount(job) facet instance, cluster_name limit 10 since 2 minutes ago
```
If any value is different from 1 then you have two or more jobs scraping the same instance in the same cluster.

https://github.com/newrelic/docs-website/pull/10553/files
### Alerts

### DIFF
--- a/alert-policies/prometheus-agent/DupMetrics.yml
+++ b/alert-policies/prometheus-agent/DupMetrics.yml
@@ -1,0 +1,27 @@
+name: Same target is scraped by different jobs
+
+description: |+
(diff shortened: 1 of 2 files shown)


josephgregoryii (commented):
Looks good at the moment - feel free to ping us in our help channel when it is ready for full review! 😄

PR #1570, opened 2022-12-15
Alert policy required data sources

brammerl:
# Summary

This PR runs a workflow that updates a submitted QuickStart's alert policy with the proper required data sources. 

## Pre merge checklist

- [ ] Did you check you NRQL syntax? - Does it work?
- [ ] Did you include an InstallPlan and Documentation reference?
- [ ] Are all documentation links publically accessible?
- [ ] Did you check your descriptive content for [voice and tone](https://docs.newrelic.com/docs/style-guide/writing-strategies/voice-strategies-docs-sound-new-relic/)? 
- [ ] Did you check your descriptive content for spelling and grammar errors?

### Dashboards 

- [ ] Does the PR contain a screenshot for each of your dashboards?
- [ ] Do your screenshots show data?
- [ ] Has the [sanitization script](https://github.com/newrelic/newrelic-quickstarts/blob/main/CONTRIBUTING.md#dashboards) been run on each dashboard?

### Alerts


utils/lib/Alert.ts
```diff
@@ -81,6 +138,76 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
+
+    if (results === undefined || results.length === 0) {
+      const error = new Error(
+        `No alert policy for quickstart ${quickstart.name} exists`
+      );
+
+      return { alertPolicy: null, errors: [error] };
+    }
+
+    const alertPoliciesWithUpdatedDataSources = results?.map(
```
zstix, line 175: At this point, I think we should be able to assume that `results` is defined and has a length.

```suggestion
    const alertPoliciesWithUpdatedDataSources = results.map(
```

utils/lib/Alert.ts
```diff
@@ -81,6 +138,76 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
+      (result: AlertPolicy) => {
+        const currDataSourceIds = result.metadata.requiredDataSources.map(
+          (dataSource) => dataSource.id
+        );
+
+        return {
+          id: result.id,
+          dataSourceIds: [
+            ...new Set([...currDataSourceIds, ...quickstart.dataSourceIds]),
```
aswanson-nr, line 184: If the quickstart has more than one data source, we have no way of telling if both data sources relate to the alert policy. So in that case, we want to **not** update the alert policy and notify ourselves that we need to do it manually.
We should probably make sure we are doing this on the dashboard side as well.

utils/lib/Alert.ts
```diff
@@ -10,6 +10,63 @@ import type {
...
+  ALERT_POLICY_REQUIRED_DATA_SOURCES_QUERY,
+  ALERT_POLICY_SET_REQUIRED_DATA_SOURCES_MUTATION,
+} from '../constants';
+
+interface RequiredDataSources {
+  id: string;
+}
+interface AlertPolicy {
+  id: string;
+  metadata: {
```
zstix, line 30: Surprisingly, `metadata` isn't actually required in our GraphQL schema. I think we should assume that it's always going to be present, but I wanted to mention it here.

We should probably update that in our schema at some point.

aswanson-nr (commented):
Notice a spot that we probably want to update around quickstarts have more than 1 data source!

zstix (commented):
A couple of small comments (plus the one Alec has). This is looking really awesome tho!!!

utils/lib/Alert.ts
brammerl: ah okay makes sense! For whatever reason, I assumed that we would just assign multiple data sources to the alert policy if there were multiple.

utils/lib/Alert.ts
```diff
@@ -81,6 +140,91 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
+    const hasFailed = quickstart.dataSourceIds.length > 1;
+
+    if (hasFailed) {
+      const error = new Error(
+        `Multiple Quickstart data sources detected for Quickstart: ${quickstart.name} with AlertPolicy: ${results[0].id} must update manually`
+      );
+      recordNerdGraphResponse(
+        hasFailed,
+        CUSTOM_EVENT.MULTIPLE_DATA_SOURCES_DETECTED
+      );
```
aswanson-nr, line 186: Could we include a bit more information here? I'm just thinking we'll want to know the quickstart's name at least so that we can go find and update the alert policy

aswanson-nr (commented):
Looks good to go to me! I had one comment about adding some extra info to the event, but that's not required at all

utils/lib/Alert.ts
brammerl: I totally forgot about this, thanks for reminding me!

utils/lib/Alert.ts
```diff
@@ -81,6 +140,92 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
+
+    if (results === undefined || results.length === 0) {
+      const error = new Error(
+        `No alert policy for quickstart ${quickstart.name} exists`
+      );
+
+      return { alertPolicy: null, errors: [error] };
+    }
+
+    const hasFailed = quickstart.dataSourceIds.length > 1;
```
caylahamann, line 177: [part 1] actually thinking about this more I feel like since `quickstart` is an input of the function, we should actually return hasFailed earlier so we don't make the request if there are multiple data sources detected for a quickstart, where i think we can put this I commented in `utils/set-alert-policy-required-datasources.ts` as part 2

utils/set-alert-policy-required-datasources.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (hasQuickstartNamesFailed) {
+    return hasQuickstartNamesFailed;
+  }
+
+  let results: SubmitSetRequiredDataSourcesMutationResult[] = [];
+
+  for (const c of chunk(Object.entries(quickstartNamesAndDataSources), 5)) {
+    const result = await Promise.all(
+      c.map(async ([_, quickstart]) => {
+        const alertPolicy = await Alert.getAlertPolicyRequiredDataSources(
```
caylahamann, line 64: [part 2] right before the request is make, we should do 
```tsx 
const hasFailed = quickstart.dataSourceIds.length > 1;
```
and then return errors before attempting to make the request

utils/lib/Alert.ts
brammerl: The reason we do it within the get alert policy function is because we need to know which alert policy is associated with that quickstart so that we know which alert policy to manually update. So I'm not sure we should move it to before we get the alert policy.

utils/newrelic/customEvent.ts
```diff
@@ -95,9 +99,10 @@ export const track = async (
...
 export const recordNerdGraphResponse = async (
   hasFailed: boolean,
-  event: CUSTOM_EVENT
+  event: CUSTOM_EVENT,
+  quickstartName?: string
 ) => {
   const status = hasFailed ? 'failed' : 'success';
 
-  await track(event, { status });
+  await track(event, { status, quickstartName });
```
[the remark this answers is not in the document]
brammerl: Not sure, we can send a conditional track depending on the presence of a quickstartName.

utils/set-alert-policy-required-datasources.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        return result;
+      })
+    );
+
+    results = [...results, ...result];
+  }
+
+  const hasFailed = results.some(
```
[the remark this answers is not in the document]
brammerl: Yeah this makes sense, my question then is what do we do with the results on line 62 from the Promise? Do we just want to run the promise and not assign any results to it?

utils/lib/Alert.ts
caylahamann: oooooooooo gotcha yes that makes sense

utils/set-alert-policy-required-datasources.ts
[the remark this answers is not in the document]
caylahamann: yes! or you could even use a reducer like this to assign the value (https://css-tricks.com/why-using-reduce-to-sequentially-resolve-promises-works/) but honestly this just a more nice to have so if it feels like it's easier to read/ work with having the returned results and then iterating over them

utils/set-alert-policy-required-datasources.ts
```diff
@@ -0,0 +1,126 @@
...
+  return hasFailed;
+};
+
+const main = async () => {
+  const [ghUrl] = passedProcessArguments();
+  const ghToken = process.env.GITHUB_TOKEN;
+
+  const hasFailed = await setAlertPoliciesRequiredDataSources(ghUrl, ghToken);
+
+  recordNerdGraphResponse(
```
caylahamann, line 109: we might need an `await` here if we want to make sure this executes before `process.exit(1)` executes

utils/set-alert-policy-required-datasources.ts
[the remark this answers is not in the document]
brammerl: I think the reason we went with this approach is because that's what happening when we assign required data sources to dashboards. Perhaps let's keep the approach of iterating over the results to maintain consistency between the two workflows.

utils/lib/Alert.ts
```diff
@@ -81,6 +140,92 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
+    if (errors) {
+      return { alertPolicy: null, errors };
+    }
+
+    if (results === undefined || results.length === 0) {
+      const error = new Error(
+        `No alert policy for quickstart ${quickstart.name} exists`
+      );
+
+      return { alertPolicy: null, errors: [error] };
```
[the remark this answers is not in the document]
brammerl: Yeah that's a good idea. I'm having it check for multiple data sources before any other conditional and then checking for the alert policy query results if there is more than 1 data source.

caylahamann (approved):

PR #1588, opened 2023-01-06
Proper queries and updated error handling for required data sources

aswanson-nr:
# Summary

We found a few issues in our approach to setting required data sources on alert policies and dashboards, this fixes those problems.


MichelLosier (approved):
Looks good :shipit:

PR #1613, opened 2023-01-30
Update primary install for "partner" data sources

zstix:
## Description
At the moment, most "partner" data sources have the same primary installation directive: the quickstart intsall nerdlet. Each data source specified the `quickstartId` for the quickstart that was most relevant (e.g. the "Akami" data source supplied the ID for the "Akamai" quickstart). Now that this work-around is no longer necessary, this PR cleans this up.

More specifically, we did the following tasks for each data source:
- Look up the "primary install" for the datasource
- If it was the quickstart install nerdlet, we found the quickstart (by ID supplied)
- We then found the installPlan for each quickstart
- We then copied the primary install directive from the install plan directly in the data source config

To support this work, I created a small script (I deemed it too much of a one-off to commit to the repository): https://gist.github.com/zstix/6f7758a7ee263e9c0ed4063c12ebfad7

## Reviewer notes

There are a few different ways these data sources had to be updated, based on the way they were configured previously and the type of primary install directive we found. In order to illustrate the decision-making process, I've attached a flowchart below.

There are a two scenarios we may want to audit at some point (outside the scope of this PR):
- When the primary install directive from the install plan was a link and it was _different_ from the existing data source fallback link. I decided to update the primary and keep the fallback as-is (since that's what the UI is doing today).
- In one instance (_MariaDB_) the primary install directive was a _targeted_ install. I wasn't sure what to do so I left it as-is (it will open the quickstart install nerdlet).

## Screenshot(s)
Screen Shot 2023-01-30 at 15 36 23

## Related Issue(s)
- `NR-67857`


josephgregoryii (approved):
Looks good! I dont remember if we want to hold off merging until the feature branch for the quickstart installation Nerdlet goes in for this ticket, but i'd double check if that is necessary.

zstix:
@josephgregoryii I'm not sure it should matter since these data sources are no longer linking to that nerdlet (well..except for MariaDB)

caylahamann (approved):

PR #1644, opened 2023-02-16
feat: Update docs to include data sources and remove installplans

brammerl:
# Summary

Updates docs to reference to data sources only. Removes all references to install plans in templates. This is to prepare for the move to only using data sources in the quickstart schema. 

###  Update 02/24 - Cayla 
I added back in the section for install plans and marked it as deprecated, additionally: 
* mapped install plan fields to data source fields 
* included note that targeted installs should be moved to core data sources and to reach out to our team 
* moved install mode section to be under data sources 
* added link to nerdgraph for available core data sources


mickeyryan42:
We should probably remove `installPlans` from _template/quickstarts/example-quickstart/config.yml

Theres also a template install.yml in the example-quickstart directory.  Other than those two things i think that covers any guides/docs where we reference install plans.

CONTRIBUTING.md
```diff
@@ -552,113 +544,12 @@ example-alert-condition2.yml
 
 For documentation on the rest of the alert condition fields, please review the [Introduction to alerts](https://docs.newrelic.com/docs/alerts-applied-intelligence/new-relic-alerts/learn-alerts/introduction-alerts)
 
-### Install plans
```
aswanson-nr, line 555: Let's not remove this, because all the related code and configurations are still living within the ecosystem. Instead we can add `Install plans (DEPRECATED)` and toss a small blurb underneath that explains the changes and points them over to data sources.

CONTRIBUTING.md
```diff
@@ -552,113 +544,12 @@ example-alert-condition2.yml
...
-They can be broken out into two categories, CORE and COMMUNITY. The CORE data sources are provided by New Relic One and do _not_ exist within this repository, the COMMUNITY data sources _are_ defined within this repository.
+They can be broken out into two categories, CORE and COMMUNITY. The CORE data sources are provided by New Relic One and do _not_ exist within this repository, the COMMUNITY data sources _are_ defined within this repository. 
+
+To see the available CORE data sources, you can use NerdGraph for to explore the available data source ids and their corresponding metadata using [this query](https://api.newrelic.com/graphiql?#query=%7B%0A%20%20actor%20%7B%0A%20%20%20%20nr1Catalog%20%7B%0A%20%20%20%20%20%20search%28filter%3A%20%7Btypes%3A%20DATA_SOURCE%7D%29%20%7B%0A%20%20%20%20%20%20%20%20results%20%7B%0A%20%20%20%20%20%20%20%20%20%20...%20on%20Nr1CatalogDataSource%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20id%0A%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A).
```
aswanson-nr, line 552: Could we make this callout again in the Quickstart `dataSourceIds` section or link to this?

CONTRIBUTING.md
aswanson-nr: I see that you've moved a lot of this content down, so I think we're covered. The only reason I could see leaving this section is to keep the docs fairly consistent, that way users don't get confused when they come back to this after these changes go in. Just something to think about, I think we're covered either way.

aswanson-nr (commented):
Some early feedback!

CONTRIBUTING.md
caylahamann: Hello!! yes it is still there, it's just moved underneath data sources because I thought it would be better to put the deprecated install plans after the data source info but can move it above again if you want!

CONTRIBUTING.md
caylahamann: heck yeah

josephgregoryii (approved):
Only some typos. Nothing to block!

zstix (approved):

PR #1649, opened 2023-02-21
Quickstart class enhancement

josephgregoryii:
## Summary

As of 02/21
* ~The updates are intended to add optional `context` to the quickstart class where `coreDataSourceIds` can be supplied to the data source class. `coreDataSourceIds` are then submitted to the new instance(s) of `Data Source`~
  *  ~Reason for this, is that we fetch core data source ids from the catalog service using an sync request.~
  * ~Since Data Source extends the `Component` class, adding an async request in the initialization of the class creates some weirdness.~
  * The idea behind this is to have a list of core data source IDs defined before initializing the classes (i.e., sending in the IDs from workflows)
* So far, this works as intended and was manually checked via PR #1639 

#### TODO:
- [x] Add unit test to Quickstart and Data sources classes to verify:
  - [x] A quickstart that has a `CORE` data source ID references does not mark either the quickstart or data source as invalid
  - [x] A stand alone data source that has a `CORE` data source ID referenced does not get marked as invalid
  - [x] Fix current tests that may be needing minor updates (this may be avoided if we ensure backwards compatibility)
- [x] (verify) If a _new_ `COMMUNITY` data source is added to the PR, both the quickstart and data source should be valid.

~While running these changes with the changes in #1639, the workflows succeed with core data source IDs. However, it fails on a new local community data source~

#### Update 02/22
Works as expected with manual test and unit tests for implementation

## New iteration
Decided to go with an approach where if a quickstart config has a core data source, we do not create a data source instance that gets generated in `this.components`

There are unit tests for that

### DIFF
--- a/utils/__tests__/check_quickstart_uniqueness.test.js
+++ b/utils/__tests__/check_quickstart_uniqueness.test.js
@@ -40,8 +40,8 @@ describe('Action: check quickstart uniqueness', () => {
   test('does not find match', () => {
     const mockGetAll = jest.fn().mockReturnValueOnce([
       { configPath: 'test/path/config.yml', config: { id: 12345 } },
-      { configPath: 'test/2/path/config.yml', config: { id: 23456} },
-      { configPath: 'test/3/path/config.yml', config: { id: 34567} },
+      { configPath: 'test/2/path/config.yml', config: { id: 23456 } },
+      { configPath: 'test/3/path/config.yml', config: { id: 34567 } },
     ]);
     Quickstart.getAll = mockGetAll;
 
@@ -54,13 +54,13 @@ describe('Action: check quickstart uniqueness', () => {
   test('finds and returns separate id matches', () => {
     const mockGetAll = jest.fn().mockReturnValueOnce([
       { configPath: 'test/path/config.yml', config: { id: 12345 } },
-      { configPath: 'test/2/path/config.yml', config: { id: 23456} },
-      { configPath: 'test/3/path/config.yml', config: { id: 12345} },
-      { configPath: 'test/4/path/config.yml', config: { id: 23456} },
-      { configPath: 'test/4/path/config.yml', config: { id: 98765} },
+      { configPath: 'test/2/path/config.yml', config: { id: 23456 } },
+      { configPath: 'test/3/path/config.yml', config: { id: 12345 } },
+      { configPath: 'test/4/path/config.yml', config: { id: 23456 } },
+      { configPath: 'test/4/path/config.yml', config: { id: 98765 } },
     ]);
 
-    Quickstart.getAll = mockGetAll
+    Quickstart.getAll = mockGetAll;
 
     checkQuickstartUniqueness();
 
--- a/utils/create_validate_pr_quickstarts.ts
+++ b/utils/create_validate_pr_quickstarts.ts
@@ -31,16 +31,29 @@ const installPlanErrorExists = (error: Error | NerdGraphError): boolean =>
   error?.extensions?.argumentPath?.includes('installPlanStepIds') &&
   error?.message?.includes('contains an install plan step that does not exist');
 
+const dataSourceErrorExists = (error: Error | NerdGraphError): boolean =>
+  'extensions' in error &&
+    error?.extensions?.argumentPath.includes('dataSourceIds') &&
+    error?.message?.includes('contains a data source that does not exist');
+
 export const countAndOutputErrors = (
   graphqlResponses: ResponseWithErrors[]
 ): number =>
   graphqlResponses.reduce((all, { errors, name }) => {
     const installPlanErrors =
       (errors?.filter(installPlanErrorExists) as NerdGraphError[]) ?? [];
+    const dataSourceErrors =
+      (errors?.filter(dataSourceErrorExists) as NerdGraphError[]) ?? [];
     const remainingErrors =
-      errors?.filter((error) => !installPlanErrorExists(error)) ?? [];
+      errors
+        ?.filter((error) => !installPlanErrorExists(error))
+        ?.filter((error) => !dataSourceErrorExists(error)) ?? [];
 
-    translateMutationErrors(remainingErrors, name, installPlanErrors);
+    translateMutationErrors(remainingErrors, name, [
+      ...installPlanErrors,
+      ...dataSourceErrors,
+    ]);
+
     return all + remainingErrors.length;
   }, 0);
 
--- a/utils/create_validate_pr_quickstarts.ts
+++ b/utils/create_validate_pr_quickstarts.ts
@@ -91,7 +91,7 @@ export const createValidateQuickstarts = async (
       if (QUICKSTART_CONFIG_REGEXP.test(filePath)) {
         return new Quickstart(
           filePath,
-          undefined,
+          undefined, // Note: We are ignoring the `basePath` argument here
           quickstartContext
         );
       }
@@ -113,7 +113,6 @@ export const createValidateQuickstarts = async (
   const invalidQuickstarts = quickstarts
     .map((qs) => {
       qs.validate();
-      console.log(qs)
       return !qs.isValid ? qs : undefined;
     })
     .filter(Boolean);
--- a/utils/create_validate_pr_quickstarts.ts
+++ b/utils/create_validate_pr_quickstarts.ts
@@ -3,7 +3,7 @@ import {
   filterOutTestFiles,
   isNotRemoved,
 } from './lib/github-api-helpers';
-import { translateMutationErrors, chunk, getPublishedDataSourceIds } from './lib/nr-graphql-helpers';
+import { translateMutationErrors, chunk } from './lib/nr-graphql-helpers';
 
 import Quickstart, { QuickstartMutationResponse } from './lib/Quickstart';
 import { CUSTOM_EVENT, recordNerdGraphResponse } from './newrelic/customEvent';
--- a/utils/create_validate_pr_quickstarts.ts
+++ b/utils/create_validate_pr_quickstarts.ts
@@ -75,22 +75,6 @@ export const createValidateQuickstarts = async (
   // Get all files from PR
   const files = await fetchPaginatedGHResults(ghUrl, ghToken);
 
-  const { coreDataSourceIds, errors } = await getPublishedDataSourceIds();
-
-  if (errors && errors.length) {
-    console.error(
-      'quickstart validate received an error fetching published data source ids'
-    );
-
-    errors.forEach((error) => {
-      console.error(error)
-    })
-
-    return false;
-  }
-
-  const quickstartContext = { coreDataSourceIds }
-
   // Get all quickstart mutation variables
   const quickstarts = filterOutTestFiles(files)
     .filter(isNotRemoved)
@@ -102,17 +86,10 @@ export const createValidateQuickstarts = async (
     )
     .flatMap((filePath) => {
       if (QUICKSTART_CONFIG_REGEXP.test(filePath)) {
-        return new Quickstart(
-          filePath,
-          undefined, // Note: We are ignoring the `basePath` argument here
-          quickstartContext
-        );
+        return new Quickstart(filePath);
       }
 
-      return getRelatedQuickstarts(
-        getComponentLocalPath(filePath),
-        quickstartContext
-      );
+      return getRelatedQuickstarts(getComponentLocalPath(filePath));
     })
     // Remove any duplicate quickstarts
     .reduce<Quickstart[]>((acc, quickstart) => {
--- a/utils/create_validate_pr_quickstarts.ts
+++ b/utils/create_validate_pr_quickstarts.ts
@@ -113,12 +113,17 @@ export const createValidateQuickstarts = async (
   const invalidQuickstarts = quickstarts
     .map((qs) => {
       qs.validate();
+      console.log(qs)
       return !qs.isValid ? qs : undefined;
     })
     .filter(Boolean);
 
   if (invalidQuickstarts.length > 0) {
-    process.exit(1);
+    if (require.main === module) {
+      process.exit(1);
+    }
+
+    return true;
   }
 
   // Submit all of the mutations in chunks of 5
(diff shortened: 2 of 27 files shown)


utils/create_validate_pr_quickstarts.ts
```diff
@@ -73,10 +89,17 @@ export const createValidateQuickstarts = async (
     )
     .flatMap((filePath) => {
       if (QUICKSTART_CONFIG_REGEXP.test(filePath)) {
-        return new Quickstart(filePath);
+        return new Quickstart(
+          filePath,
+          undefined,
```
josephgregoryii, line 94: A comment here may be worth putting - we're basically "ignoring" the `basePath` argument ehre

utils/__tests__/check_quickstart_uniqueness.test.js
```diff
@@ -40,8 +40,8 @@ describe('Action: check quickstart uniqueness', () => {
   test('does not find match', () => {
```
josephgregoryii, line 40: I guess formatter hasnt been run on our repo in a while. Sorry for the extra 'fluff'

You can assume anything that doesn't touch a `quickstart` or `data source` can be ignored

utils/lib/DataSource.ts
```diff
@@ -23,13 +24,42 @@ export interface DataSourceMutationResponse {
   };
 }
 
+interface DataSourceContext extends QuickstartContext {}
```
josephgregoryii, line 27: this might not be necessary. Wanted to keep the namespace consistent.

utils/lib/helpers.ts
```diff
@@ -27,9 +28,10 @@ export const prop =
  * @returns quickstarts that use the passed in component
  */
 export const getRelatedQuickstarts = (
-  localComponentPath: string
+  localComponentPath: string,
+  context?: QuickstartContext
 ): Quickstart[] =>
-  Quickstart.getAll().filter((q) =>
+  Quickstart.getAll(undefined, context ?? {}).filter((q) =>
```
mickeyryan42, line 34: Can we/should we default the context on the static method to {} also instead of making it the callers responsibility?

mickeyryan42 (commented):
Nice! Just a few non-blocking comments/questions.  I wonder too if since `create_validate_pr_quickstarts` is now calling `getPublishedDataSourceIds` whether we should have a test that mocks an error response and validates the error handling.

utils/lib/__tests__/DataSource.test.js
```diff
@@ -16,8 +16,8 @@ jest.mock('fs', () => {
   };
 });
 
-jest.spyOn(global.console, 'log').mockImplementation(() => {});
-jest.spyOn(global.console, 'error').mockImplementation(() => {});
```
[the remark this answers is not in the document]
josephgregoryii: Meant to put it back in, will uncomment those

utils/lib/helpers.ts
josephgregoryii: That seems like a much better approach.

utils/lib/DataSource.ts
aswanson-nr: Would it make more sense to have the QuickstartContext extend the DataSourceContext? That way the data source can define new context parameters without needed to modify the quickstart side of the things.

utils/lib/DataSource.ts
```diff
@@ -23,13 +24,42 @@ export interface DataSourceMutationResponse {
...
+
   /**
    * @returns Filepath for the configuration file (from top-level directory).
    */
   getConfigFilePath() {
     const id = this.identifier;
 
+    if (this.isCoreDataSource) {
+      return id;
+    }
```
aswanson-nr, line 61: This may be something we do elsewhere, but I think this makes things a little confusing. This function is `getConfigFilePath` but if that data source is CORE, then it returns an ID. We're kinda abusing the fact that this function returns a string. I could see it being far more clear if it returned the empty string in this case. Let me know what you think, just spit balling here.

utils/lib/DataSource.ts
```diff
@@ -70,6 +100,10 @@ class DataSource extends Component<DataSourceConfig, string> {
   }
 
   getConfigContent() {
+    if (this.isCoreDataSource) {
+      return { id: this.identifier } as DataSourceConfig;
+    }
```
aswanson-nr, line 105: Same comment here as the `getConfigFilePath` method. 
It seems like we're doing gymnastics just to let the DataSource class support being a "core" data source. Originally, these classes we're created to represent the physical files in the repository, so I'm wondering if it would be better to handle the use of core data source entirely at the quickstart level and not even create a DataSource instance for them.

utils/lib/Quickstart.ts
```diff
@@ -256,16 +268,17 @@ class Quickstart {
    * Static method that returns a list of every quickstarts
    * @returns - A list of all quickstarts
    */
-  static getAll(basePath?: string): Quickstart[] {
+  static getAll(basePath?: string, context?: QuickstartContext ): Quickstart[] {
```
aswanson-nr, line 271: Two optional parameters in a row leads to some weird call signatures when you're calling the function, such as `getAll(undefined, {})` which is really confusing without hover docs. Since these are both optional, why don't we convert it to an `options` object argument with the fields `basePath` and `context`

aswanson-nr (commented):
Overall looks like a good direction! I added few comments around the class structure, let me know what you think.

utils/lib/DataSource.ts
josephgregoryii: Ah, that makes more sense.

mickeyryan42 (dismissed):
Realized i didnt approve but just commented.  approving now so it can get through!

utils/lib/DataSource.ts
josephgregoryii: Yeah, thats a great callout. 

I kept that portion in before "revalidating" the data source in the constructor (lines 39-52). Sincethe quickstart class only checks to see if the data source `isValid`.

`getConfigFilePath` doesn't actually get re-ran, so this block of code _never_ gets executed. 

Removed!

utils/lib/DataSource.ts
josephgregoryii: I agree with this, since `core` data sources don't even really live in this repository.

I think it would be better to handle core data sources entirely at the quickstart level, but that would add some weirdness into the generalized nature of the class (`getComponents` for example)

I think it may be out of scope for this ticket, but happy to discuss further or do a follow-up ticket to help move along IP -> DS

josephgregoryii: To add on from something @MichelLosier and i discussed when we originally were brainstorming on this (sorry for paraphrasing, its been a little bit since the initial convo) -

We kept coming back to the idea that a Data Source should be seperated from the `Component` structure and have it live individually _along_ the quickstart. Since a quickstart config only cares about the ID of a data source for mutation purposes.

utils/lib/DataSource.ts
```diff
@@ -23,7 +23,39 @@ export interface DataSourceMutationResponse {
...
+
 class DataSource extends Component<DataSourceConfig, string> {
+  private isCoreDataSource: boolean = false;
+
+  constructor(
+    identifier: string,
+    basePath: string = path.join(__dirname, '..', '..'),
+    context: DataSourceContext = {}
+  ) {
+    super(identifier, basePath);
```
zstix, line 38: As you have pointed out below, `super` is going to call `getConfigFilePath()` and `getConfigContent`. Both have their own calls to helper methods like `removeBasePath` and node libraries like `path.dirname`, `glob.sync`, etc. I'm not 100% sure, but it's very likely that some of these functions might throw errors if we're giving them content we _know_ is invalid (no config files). That could cause some weird things like failing jobs or a bunch of static in our outputs.

We might want to revist how we approach this. As a stop-gap, we _could_ overwrite those top-level methods to be no-ops in this class...but that feels like a stop-gap not a proper solution.

utils/lib/DataSource.ts
```diff
@@ -23,7 +23,39 @@ export interface DataSourceMutationResponse {
...
+      this.identifier = identifier;
+      this.isValid = true;
+
+      // Since we know this is a CORE data source,
+      // when a quickstart submits a mutation, it relys on the
+      // data source component to supply the 'ID' of the data source.
+
+      // The data source doesn't live in this repository, so we can
+      // give supply the mutation variables what it wants (the data source ID)
+      this.config = { id: this.identifier } as DataSourceConfig;
```
zstix, line 55: `this.config` in this case is _not_ actually a `DataSourceConfig`. We can cast it with `as` but the underlying data doesn't match this at all. This means that any type checking on core data sources is going to be impaired.

As I type this comment (along with the one above), I feel more sure that we should be looking for another solution. Perhaps we can address the community / core data source problem at script time by segmenting them out. Then we could have totally separate classes (or collection of functions) to handle these totally separate situations.

zstix (changes requested):
Two comments about the direction. Full disclosure, I stopped reviewing after the `DataSource.ts` file due to getting pulled away from my computer, but I think those two comments probably inform next steps (from my perspective).

josephgregoryii:
Putting this back into draft to come up with a better solution

utils/lib/DataSource.ts
aswanson-nr: A possible idea:
Does `getComponents` even need to return the core data sources? Could they live separately from the other "components" in the Quickstart class?

utils/lib/InstallPlan.ts
```diff
@@ -90,7 +90,7 @@ class InstallPlan extends Component<InstallPlanConfig, string> {
   public async submitMutation(dryRun = true) {
     if (!this.isValid) {
       console.error(
-        `Install plan is invalid.\nPlease check the dashboard at ${this.identifier}\n`
+        `Install plan is invalid.\nPlease check the install plan at ${this.identifier}\n`
```
josephgregoryii, line 93: Noticed a typo here

utils/lib/Quickstart.ts
```diff
@@ -42,36 +42,57 @@ const SUPPORT_LEVEL_ENUMS: SupportLevelMap = {
...
+};
 
 const ConfigToMutation: ConfigToMutationMap[] = [
-  { configKey: 'alertPolicies', mutationKey: 'alertConditions', ctor: Alert },
-  { configKey: 'dashboards', mutationKey: 'dashboards', ctor: Dashboard },
   {
-    configKey: 'dataSourceIds',
+    configKey: ConfigKey.AlertPolicies,
+    mutationKey: 'alertConditions',
+    ctor: Alert,
```
aswanson-nr, line 61: This is just the constructor, right? Could we just rename it to `constructor`, `ctor` feels needlessly cryptic

utils/lib/Quickstart.ts
```diff
@@ -42,36 +42,57 @@ const SUPPORT_LEVEL_ENUMS: SupportLevelMap = {
...
+    ctor: Dashboard,
+  },
+  {
+    configKey: ConfigKey.DataSource,
     mutationKey: 'dataSourceIds',
     ctor: DataSource,
   },
 ];
 
+export interface QuickstartContext {
```
aswanson-nr, line 75: Maybe a comment here acknowledging the way we're handling core data sources. That way we have context during the next change.

utils/create_validate_pr_quickstarts.ts
```diff
@@ -73,10 +102,17 @@ export const createValidateQuickstarts = async (
     )
     .flatMap((filePath) => {
       if (QUICKSTART_CONFIG_REGEXP.test(filePath)) {
-        return new Quickstart(filePath);
+        return new Quickstart(
+          filePath,
+          undefined, // Note: We are ignoring the `basePath` argument here
```
aswanson-nr, line 107: why are we ignoring the `basePath` argument here?

aswanson-nr: Thinking about this more, I could see making the `basePath` and `quickstartContext` arguments an `options` object since they are both optional.

aswanson-nr (commented):
I think this approach cleans things up a lot!

utils/lib/Quickstart.ts
josephgregoryii: `ctor` is what we had originally, i think there was a reason for this but im happy to address it in this PR

utils/create_validate_pr_quickstarts.ts
josephgregoryii: > why are we ignoring the `basePath` argument here?
It was just a convenience thing to rely on the default. 

I think an options object is a good way to go. Consider it changed!

josephgregoryii: Addresed in commit https://github.com/newrelic/newrelic-quickstarts/pull/1649/commits/fbde806738e80eca9a0ff4b8615fb0c92d70e3a5

utils/lib/Quickstart.ts
josephgregoryii: https://github.com/newrelic/newrelic-quickstarts/pull/1649/commits/966191cc9c4f208ad3adf5eb4ad8086e81501a93

utils/lib/Quickstart.ts
josephgregoryii: I guess my commit messaged left the chat

https://github.com/newrelic/newrelic-quickstarts/pull/1649/commits/549921f036735f483af7de61afe7de040101fce2

zstix:
Screen Shot 2023-02-24 at 15 24 54

utils/lib/Quickstart.ts
```diff
@@ -42,19 +42,28 @@ const SUPPORT_LEVEL_ENUMS: SupportLevelMap = {
 type ComponentType = typeof Alert | typeof Dashboard | typeof DataSource;
 type Components = InstanceType<ComponentType>;
 
+enum ConfigKey {
+  DataSource = 'dataSourceIds',
```
aswanson-nr, line 46:
```suggestion
  DataSource = 'dataSourceIds',
```

aswanson-nr (approved):

zstix (approved):

PR #1655, opened 2023-02-22
Script to create a new data sources from install plans

zstix:
## To do
- [x] Find quickstart that "matches" each install plan we're converting (currently based on `title`)
- [x] (in progress) For each match, use that icon as the data source icon
- [ ] (optional) Create a generic icon for the data sources that don't have a match
- [x] (optional) Ability to load a list of install plan IDs from a file


MichelLosier (approved):
Looks great!

PR #1662, opened 2023-02-24
Add script to assign known datasources to quickstarts from install plans

MichelLosier:
# Summary

This PR adds a script that updates quickstarts with data sources, based on a mapping of install plans to their best matched data source. This mapping is provided as a CSV.

A side effect of serializing in and out of yaml is we lose comments, these comments mostly are just field descriptions copied over from the quickstart yaml template. Only 85 / 155 quickstart config yaml had this commenting.

To run:

From `./utils`
```
yarn run ts-node ./migration_scripts/assign-data-sources-to-quickstarts.ts ../../../relative-path-to-your.csv
```


utils/migration_scripts/assign-data-sources-to-quickstarts.ts
```diff
@@ -0,0 +1,117 @@
...
+          return 1;
+        } else {
+          return 0;
+        }
+      },
+    };
+
+    fs.writeFileSync(
+      path.resolve(quickstart.basePath, quickstart.identifier),
+      yaml.dump(qsYaml, yamlOptions)
```
aswanson-nr, line 106: My only concern here is we've seen `yaml.dump` outputting `description: >` instead of the pipe operator `description: |`. This is important because `>` turns newlines into spaces and `|` keeps the formatting as-is, which we need because markdown formatting relies on the whitespace. So we should just verify that before merging!

aswanson-nr (commented):
I think this is ready to go, just not approving till we check my one comment!

utils/migration_scripts/assign-data-sources-to-quickstarts.ts
mickeyryan42: I'm going to go check on this again to be sure, but I saw using the lineWidth: -1 option actually outputs the `|` as we want which is one reason i used it, as well as just wanting the unlimited line widths.  Going to make sure thats the case!

mickeyryan42: Confirmed! Using lineWidth it generates  `|` , taking it out it generates the `>`.  I wish we'd known that sooner, that was annoying in the past getting the `>`! Going to approve this as well now

mickeyryan42 (approved):
Love the sorting option 👍

PR #1665, opened 2023-02-27
feat(pihole): Added a quickstart for pihole

LeonAdato:
# Summary

This adds a dashboard and quickstart for pihole monitoring data, which leverages the pihole_detailed.yml integration on https://github.com/newrelic/nri-flex/pull/.

## Pre merge checklist

- [ Y ] Did you check you NRQL syntax? - Does it work?
- [ Y  ] Did you include a Data source and Documentation reference?
- [ Y ] Are all documentation links publicly accessible?
- [ Y ] Did you check your descriptive content for [voice and tone](https://docs.newrelic.com/docs/style-guide/writing-strategies/voice-strategies-docs-sound-new-relic/)?
- [ Y ] Did you check your descriptive content for spelling and grammar errors?
- [ Y ] Did you review your content with a subject matter expert? (e.g. a Browser agent quickstart is reviewed with a member of the Browser Agent team)

### Dashboards

- [ Y ] Does the PR contain a screenshot for each of your dashboards?
- [ Y ] Do your screenshots show data?
- [ Y ] Has the [sanitization script](https://github.com/newrelic/newrelic-quickstarts/blob/main/CONTRIBUTING.md#dashboards) been run on each dashboard?

### Alerts

- [ n/a ] Did you check that your alerts actually work?


data-sources/pihole/config.yml
```diff
@@ -0,0 +1,22 @@
+id: pihole
+displayName: PiHole
+description: |
+A [PiHole](https://pihole.net) is a free DNS server utility that runs on a Raspberry Pi (or other inexpensive device) and captures DNS requests before they're sent out of the network, filtering discrete elements that are spam, harmful, or wasteful. The result is a faster and safer network. 
+
+This quickstart monitors the PiHole by leveraging it's built-in API to display information about the type and quantify of network requests being handled. 
+
+This quickstart requires the pihole-detailed.yml file found at [https://github.com/newrelic/nri-flex/tree/master/examples](https://github.com/newrelic/nri-flex/tree/master/examples).
+
+icon: logo.svg
```
zstix, line 10: I need to take a look into why this line is causing a linting error. In the meantime, I _did_ notice that the SVG file supplied is not valid (Github can't even render it). Would you be willing to update and submit a new version of that logo?

zstix: It looks like the icon provided for the quickstart is valid, so perhaps we can use that one (if the content is the same).

zstix: Ah, I believe the linting error is due to your multi-line description above _not_ being indented in. The linter thinks that entire bit of text is the key. If you update that, this should be happy.

zstix (changes requested):
There's a couple small formatting issues that is preventing this from linting / building successfully. Once those are addressed this should be good to go!

data-sources/pihole/config.yml
LeonAdato: I didn't include a logo at all, and presumed a default would be used (I'm a NR employee, so I didn't think a logo was needed. 

I'll update the text. Just a sec...

LeonAdato: OK, updated. I'm VERY new to git/github so if I have to do something besides hit the "resolve conversation" button below, lemme know. Either here or in slack.

data-sources/pihole/config.yml
```diff
@@ -0,0 +1,22 @@
...
+  primary:
+    link:
+      url: https://www.newrelic.com
+keywords:
+  - pihole
+  - dns
+  - flex
+  - api
+categoryTerms:
+  - dns
```
zstix, line 21: One small adjustment: `dns` is not a valid category (at the moment).

```suggestion
```

zstix (approved):

PR #1672, opened 2023-03-01
Stop hiding errors

aswanson-nr:
# Summary
* fixes some code that was dropping errors


utils/lib/nr-graphql-helpers.ts
```diff
@@ -114,6 +114,10 @@ export const translateMutationErrors = (
     } else {
       console.error(`- ${error.message}`);
     }
+
+    if (error.message) {
+      console.error(error.message);
+    }
```
zstix, line 120: This looks like it doubles our error logging. We probably don't need to do this.

```suggestion
```

zstix (approved):
:shipit:

PR #1674, opened 2023-03-01
chore: Add missing icons to datasources

MichelLosier:
Datasources require icons, this PR adds icons for the datasources that are missing them.


data-sources/ansible-automation-controller/config.yml
```diff
@@ -6,3 +6,4 @@ install:
   primary:
     link:
       url: https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/ansible-controller-integration/
+icon: logo.png
```
aswanson-nr, line 9: This logo is going to look terrible when it's shrunk down

aswanson-nr (approved):
Noticed one that is not going to look good at smaller sizes, but the rest look great!

data-sources/ansible-automation-controller/config.yml
MichelLosier: Good call, updated the icon

aswanson-nr (approved):

PR #1702, opened 2023-03-16
Update github actions to remove set-output usage

mickeyryan42:
https://issues.newrelic.com/browse/NR-88402

In typescript files, the github doc recommends upgrading @actions/core to v1.10.0 and using the toolkit functions directly to set environment variables via core.setOutput.  This is much better than what i originally did, which was to change the return of the scripts to output the values we need.

### DIFF
--- a/.github/workflows/preview-links.yml
+++ b/.github/workflows/preview-links.yml
@@ -34,9 +34,8 @@ jobs:
         env:
           PR_URL: ${{ github.event.pull_request.url }}
         run: |
-          COMMENT = yarn create-preview-links "$PR_URL" "$PR_NUMBER"
-          echo "COMMENT=$COMMENT" >> $GITHUB_ENV
+          yarn create-preview-links "$PR_URL" "$PR_NUMBER"
       - name: Comment
-        if: env.COMMENT != ''
+        if: env.comment != ''
         run: |
-          gh pr comment "$PR_NUMBER" --body "env.COMMENT"
+          gh pr comment "$PR_NUMBER" --body "env.comment"
--- a/.github/workflows/preview-links.yml
+++ b/.github/workflows/preview-links.yml
@@ -34,9 +34,9 @@ jobs:
         env:
           PR_URL: ${{ github.event.pull_request.url }}
         run: |
-          comment = yarn create-preview-links "$PR_URL" "$PR_NUMBER"
-          echo "comment=$comment" >> $GITHUB_ENV
+          COMMENT = yarn create-preview-links "$PR_URL" "$PR_NUMBER"
+          echo "COMMENT=$COMMENT" >> $GITHUB_ENV
       - name: Comment
-        if: env.comment != ''
+        if: env.COMMENT != ''
         run: |
-          gh pr comment "$PR_NUMBER" --body "env.comment"
+          gh pr comment "$PR_NUMBER" --body "env.COMMENT"
--- a/.github/workflows/preview-links.yml
+++ b/.github/workflows/preview-links.yml
@@ -36,8 +36,8 @@ jobs:
         run: |
           yarn create-preview-links "$PR_URL" "$PR_NUMBER"
       - name: Comment
-        if: env.comment != ''
+        if: steps.links.outputs.comment != ''
         env:
-          COMMENT: ${{ env.comment }}
+          COMMENT: ${{ steps.links.outputs.comment }}
         run: |
           gh pr comment "$PR_NUMBER" --body "$COMMENT"
--- a/.github/workflows/preview-links.yml
+++ b/.github/workflows/preview-links.yml
@@ -34,11 +34,9 @@ jobs:
         env:
           PR_URL: ${{ github.event.pull_request.url }}
         run: |
-          yarn create-preview-links "$PR_URL" "$PR_NUMBER"
-
+          comment = yarn create-preview-links "$PR_URL" "$PR_NUMBER"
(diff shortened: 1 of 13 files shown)


utils/__tests__/create-preview-links.test.ts
```diff
@@ -151,9 +151,6 @@ describe('create-preview-links', () => {
 
       const res = await generatePreviewComment('testurl', '1', 'testtoken');
       expect(res).toBe(true);
-      expect(console.log).toHaveBeenCalledWith(
```
aswanson-nr, line 154: Could we check for the call to `core.setOutput` instead?

utils/__tests__/dashboard-helper.test.js
```diff
@@ -108,10 +108,6 @@ describe('dashboard-helper', () => {
         'raw-url/dashboards/cool-dash/cool-dash.json',
         { headers: { authorization: 'token token' } }
       );
-      expect(console.log.mock.lastCall[0]).toContain(
```
aswanson-nr, line 111: Same comment as the `create-preview-links.test.ts` comment

aswanson-nr (commented):
This is looking good to me! If we can verify that the `create-preview-links` workflow still works correctly, than I think we can say this is good to go.

mickeyryan42:
> This is looking good to me! If we can verify that the `create-preview-links` workflow still works correctly, then I think we can say this is good to go.

@aswanson-nr It appears to, if you see this pr here i made to test on my fork it added comments for apache which i added to test.  And also added openai which i guess was because my main branch wasnt updated with that addition!
https://github.com/mickeyryan42/newrelic-quickstarts/pull/16

utils/__tests__/dashboard-helper.test.js
mickeyryan42: ah, i did mean to do this!  will update and let you know once done

mickeyryan42: done! should be good to go!

.github/workflows/preview-links.yml
```diff
@@ -35,10 +35,7 @@ jobs:
...
           yarn create-preview-links "$PR_URL" "$PR_NUMBER"
-
       - name: Comment
-        if: steps.links.outputs.comment != ''
-        env:
-          COMMENT: ${{ steps.links.outputs.comment }}
+        if: env.comment != ''
         run: |
-          gh pr comment "$PR_NUMBER" --body "$COMMENT"
+          gh pr comment "$PR_NUMBER" --body "env.comment"
```
aswanson-nr, line 41: I'd be a little surprised if this is working as is, [the docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-writing-an-environment-variable-to-github_env), show that you need to do a "${{ }}" to access the values.
We'll also want to pull this value into a string variable before using it directly in the script. See my other comment in the `validate_packs.yml` file

.github/workflows/validate_packs.yml
```diff
@@ -352,12 +352,10 @@ jobs:
...
       - name: Comment
-        if: steps.helper.outputs.comment != ''
+        if: env.comment != ''
         env:
           PR_NUMBER: ${{ steps.get_pr_number.outputs.pr-number }}
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          COMMENT: ${{ steps.helper.outputs.comment }}
         run: |
-          gh pr comment "$PR_NUMBER" --body "$COMMENT"
+          gh pr comment "$PR_NUMBER" --body "env.comment"
```
aswanson-nr, line 361: To keep this secure and not allow any injection attacks, let's keep the env declaration. 
```
env:
  PR_NUMBER: ${{ steps.get_pr_number.outputs.pr-number }}
  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  COMMENT: ${{ env.comment }}
run: |
  gh pr comment "$PR_NUMBER" --body "$COMMENT"
```
This will help us avoid being vulnerable to things like [word splitting](https://github.com/koalaman/shellcheck/wiki/SC2086)

aswanson-nr (changes requested):
Found two areas that we'll need to clean up. @mickeyryan42 The reason you're seeing it work correctly on your fork is because the new workflows have not been merged in the main branch, so you're actually running the current `main` workflows, not the updates you just made.

.github/workflows/preview-links.yml
mickeyryan42: I saw that in the docs and was a little curious, because we werent using it in the `gh pr comment` command in that way.  i see now though, its because its a string and not env var! i updated this and `validate_packs.yml` to keep the env declaration and use the string we set

d3caf (commented):
I'm getting `invalid refspec '+refs/pull//merge:refs/remotes/pull//merge'` on all the `validate-*` workflows. I tested against main and didn't have the same issue so I'm not sure if it's something to do with these changes.

https://github.com/d3caf/newrelic-quickstarts/actions/runs/4448797645

mickeyryan42:
> I'm getting `invalid refspec '+refs/pull//merge:refs/remotes/pull//merge'` on all the `validate-*` workflows. I tested against main and didn't have the same issue so I'm not sure if it's something to do with these changes.
> 
> https://github.com/d3caf/newrelic-quickstarts/actions/runs/4448797645

@d3caf This should now be fixed!

d3caf (approved):
NOICE

PR #1709, opened 2023-03-17
Extra logging idea (innovation)

aswanson-nr:
# Summary
We've said time and time again that we need more logs during normal workflow runs. I decided to take a crack at this problem with my innovation time. This PR adds a logging module and more logging statements to our various workflows and utility functions. It's very much a first pass and I would love to discuss it more with the team and get it into a state we all like!


MichelLosier (approved):
Looks great! Thanks for taking the time to make this improvement!

utils/lib/github-api-helpers.ts
```diff
@@ -51,23 +52,31 @@ export const fetchPaginatedGHResults = async (
...
+      // TODO: this should happen after the resp.ok check
       const responseJson = await resp.json();
 
       if (!resp.ok) {
+        logger.error(`Error from Github API`, {
+          url: nextPageLink,
+          status: resp.statusText,
+          code: resp.status,
+        });
         throw new Error(`Github API returned: ${responseJson.message}`);
```
josephgregoryii, line 73: Would we want to also pass along the json response to the logger?

josephgregoryii (approved):
What a _wonderful_ innovation! Think its a great first pass.

I think just some overall question(s) below.

utils/lib/github-api-helpers.ts
josephgregoryii: jk, i realize that since it throws the error it'll display that in the workflows

PR #1737, opened 2023-04-10
Azure monitor

jcountsNR:
New data source, Azure Monitor. To be used by Product Partnerships team for Azure monitor integrations.


data-sources/azure-monitor/config.yml
```diff
@@ -0,0 +1,9 @@
+id: azure-monitor
```
aswanson-nr, line 1: We already have an Azure data source, could we possibly use that?

aswanson-nr (commented):
Hey @jcountsNR,
There is already a core Azure data source in the system that links out to the setup screen for the Azure integration in the product. 
Could we make use of that data source instead of creating a new one?

Possible issues:
- Existing Azure data sources opens a new tab, which will kick you out of the quickstart install process

data-sources/azure-monitor/config.yml
aswanson-nr: That data source is a different method for reporting Azure data, this is a new way.

PR #1831, opened 2023-06-07
Confluent Cloud dashboard update

4demos:
# Summary

update to Confluent Cloud dashboard

### Dashboards

- [x] Does the PR contain a screenshot for each of your dashboards?
- [x] Do your screenshots show data?
- [x] Has the [sanitization script](https://github.com/newrelic/newrelic-quickstarts/blob/main/CONTRIBUTING.md#dashboards) been run on each dashboard?


caylahamann (approved):

4demos:
"ERROR: The following errors occurred while validating: quickstarts/confluent-cloud/config.yml
- quickstartMetadata/dashboards/rawConfiguration/variables/nrqlQuery/accountIds: `accountIds` can't be blank" 
accountIds not used in confluent-cloud.json, they all replaced with accountId and set to 0

dashboards/confluent-cloud/confluent-cloud.json
```diff
@@ -1,622 +1,633 @@
...
+      ]
+    }
+  ],
+  "variables": [
+    {
+      "name": "kafka_id",
+      "items": null,
+      "defaultValues": [],
+      "nrqlQuery": {
+        "accountId": 0,
```
aswanson-nr, line 624: Hey @4demos, it looks like this is the culprit. If we convert it over to `accountIds` version it should work just fine.
```suggestion
        "accountIds": [],
```

aswanson-nr (changes requested):
Hey @4demos, Figured out where that workflow was failing!

dashboards/confluent-cloud/confluent-cloud.json
4demos: good catch! 
still not sure why in most cases we replace `"accountIds": [...]` with `"accountId": 0` but here we should keep "accountIds"

brammerl (approved):

dashboards/confluent-cloud/confluent-cloud.json
aswanson-nr: Agreed! Did the dashboard cleanup script adjust it to `"accountId": 0`?

aswanson-nr (approved):

dashboards/confluent-cloud/confluent-cloud.json
4demos: `sanitize-dashboard` doesn't set "accountId": 0, just removes the actual account. But instructions https://developer.newrelic.com/contribute-to-quickstarts/build-a-quickstart/add-dashboard/ tell submitters: "Set your accountId to 0". If it's not a requirement any longer and `"accountIds": []` is fine, may be it's time to update the docs?

aswanson-nr: @4demos, thank you! I'll get the docs updated

PR #1856, opened 2023-06-20
fix(Puppet install): Updated the content

NSSPKrishna:
# Summary
The document has been revised to direct customers to the actively maintained and supported resource.
Related jira: https://issues.newrelic.com/browse/NR-127693


mickeyryan42 (approved):
lgtm 👍

PR #1875, opened 2023-07-10
Deprecate "local quickstart preview" docs

d3caf:
It seems that viewing quickstarts locally is no longer possible via the steps we noted in the docs. Updated the docs to reflect this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -684,41 +684,45 @@ icon.png
   ```
 
 ## Quickstart Preview
+⚠️  There is currently no way to preview Quickstarts locally as Jumpstart no longer owns the public I/O website (newrelic.io).
 
-Quickstart Previews are available for contributors to review their new or improved quickstarts directly from the Public I/O site! We provide two ways to view a preview:
+<details>
+  <summary>Deprecated preview steps</summary>
+    Quickstart Previews are available for contributors to review their new or improved quickstarts directly from the Public I/O site! We provide two ways to view a preview:
 
-### Local Quickstart Preview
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -684,41 +684,45 @@ icon.png
   ```
 
 ## Quickstart Preview
+⚠️  There is currently no way to preview Quickstarts locally as Jumpstart no longer owns the public I/O website (newrelic.io).
```
zstix, line 687: I think the issue with the preview not working stems from our new tech stack (Drupal) vs the Gatsby-based approach we had before. Furthermore, because this is a public-facing repository, including our team name might not be helpful to the general public (what's a JumpStart?)

quickstarts/azure/azure-functions-workflow/config.yml
```diff
@@ -1,3 +1,4 @@
+id: 099258cc-d981-40d4-907c-3b2881595735
```
zstix, line 1: Was this addition intentional?

CONTRIBUTING.md
d3caf: Yeah, it apparently used to work? Unless you mean that we recently switched to Drupal (I have no idea).

Good point on the team name!

quickstarts/azure/azure-functions-workflow/config.yml
d3caf: Nope. No idea where that came from or why my commit message is incorrect... I'll fix it.

zstix (approved):

PR #1888, opened 2023-07-24
Dashboard validation script handles errors across multiple lines

sarahkitten:
https://issues.newrelic.com/browse/NR-120279

Updates to dashboard validation script and unit tests.

Switches from parsing the json with json.stringify to neatJSON in order to account for multi-line errors - neatJSON condenses multi-line errors (like the one mentioned in the linked ticket) into one line before the line is checked.

This also means that line numbers are no longer accurate, so I removed them from the warning message.

Updated unit tests to use `getWarnings` and added a test for errors across multiple lines


utils/dashboard-helper.ts
```diff
@@ -62,7 +78,7 @@ export const runHelper = async (
     return false;
   }
 
-  const warnings: string[] = [];
+  let warnings: string[] = [];
```
MichelLosier, line 81: We can still use `const` here since we are not replacing the reference, just adding entries with `push` as before

utils/__tests__/dashboard-helper.test.js
```diff
@@ -105,14 +116,15 @@ describe('dashboard-helper', () => {
       });
 
       const output = await runHelper('http://localhost', 'token');
+      console.debug('output: ', output)
```
MichelLosier, line 119: Do we still want this console.debug here? or do think its necessary to understand the test

utils/package.json
```diff
@@ -45,13 +45,13 @@
     "graphql-markdown": "^6.0.0",
     "jest": "^27.5.1",
     "js-yaml": "^4.1.0",
+    "neatjson": "^0.10.6",
```
MichelLosier, line 48: It looks a fair bit of the CI checks are failing because this new dependency is not in the yarn.lock file. 

```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`
```

Lets make sure this dependency was installed through yarn

utils/dashboard-helper.ts
```diff
@@ -81,16 +97,10 @@ export const runHelper = async (
...
-
-      dashLines.forEach((line, lineNumber) => {
-        const output = checkLine(line);
-        if (output.length > 0) {
-          output.forEach((o) =>
-            warnings.push(`| ${o} | ${dash.filename} | ${lineNumber + 1} |`)
-          );
-        }
-      });
+      const output = getWarnings(responseJSON);
```
MichelLosier, line 100: [nit] I kind of expect what we get back from `getWarnings` to be called `warnings` and maybe how we process them later on `102` as being the formatted `output`

MichelLosier (changes requested):
Couple things to look at here. 

I don't know if its the scope of this ticket, but maybe a follow up one to consider is not validating by checking raw json line by line with regex - but actually walking the tree of the json object and performing validation. Otherwise we are in this situation of maintaining regex rules to infer the deserialized values to check against - if we are doing that we might as well just deserialize the object.

utils/__tests__/dashboard-helper.test.js
```diff
@@ -20,51 +22,60 @@ jest.mock('../lib/github-api-helpers', () => {
...
+      expect(getWarnings({"accountIds":[]})).toHaveLength(0);
     });
+
+    test('finds warnings across multiple lines', () => {
+      const multiLine = {
+        "accountIds": [ 
+          12345678 
+        ],
+      }
+      expect(getWarnings(multiLine)).toHaveLength(1);
```
caylahamann, line 72: do we know if the `getWarnings` returns the `multiLine` object with line breaks??

caylahamann: jk i'm seeing that it's a wrapper around `neatJSON` - what does `neatJSON` do?

sarahkitten: It's like a more powerful alternative to JSON.stringify that I'm using to collapse unnecessary line breaks so that multi-line accountIds (for example) end up as one line to check

MichelLosier (approved):

utils/dashboard-helper.ts
```diff
@@ -28,14 +29,29 @@ export const checkLine = (line: string) => {
   return warningsFound;
 };
 
+export const getWarnings = (dashboardJson: any) => {
+  const dashLines = neatJSON(dashboardJson, {padding:2}).split('\n');
```
aswanson-nr, line 33: I'd rather not pull in a new package (especially one that isn't at a 1.0 state yet) just to do some JSON handling. If all we're using is the ability to one-line the array values, then the following snippet might work just fine:
```js
JSON.stringify(dashboardJson, (k, v) => {
  if (Array.isArray(v)) {
    return JSON.stringify(v)
  }
  return v
}, 2);
```

This snippet is runnable and you can see that the output puts the array on a single line:
```js
console.log(JSON.stringify({id: [2], desc: 'asdfadsfa'}, (k, v) => {
  if (Array.isArray(v)) {
    return JSON.stringify(v)
  }
  return v
}, 2));
```

aswanson-nr (changes requested):

utils/dashboard-helper.ts
sarahkitten: Took your suggestion!

MichelLosier (approved):

aswanson-nr (approved):

PR #1923, opened 2023-08-11
chore: update featured list

jcountsNR:
Adding and removing a few IO listings from featured list.


quickstarts/codestream/codestream/config.yml
```diff
@@ -40,4 +40,3 @@ documentation:
 website: https://codestream.com
 # Keywords for filtering / searching criteria in the UI
 keywords:
```
brammerl, line 42: If there aren't any keywords I think you may have to delete the field all together.

louisleungk: maybe we can add a tag for "IDE"?

brammerl:
@louisleungk requested we add `featured` keyword to docker otel quickstart so i went ahead and did that.

seemantk:
Thank you @brammerl

seemantk (dismissed):

quickstarts/codestream/codestream/config.yml
brammerl: @seemantk what do you think of this suggestion? Seems fair to me.

brammerl (approved):

PR #2000, opened 2023-08-30
Return quickstart keywords as empty array if no keywords are provided

sarahkitten:
https://issues.newrelic.com/browse/NR-154816

- [x] When a quickstart config does not include any keywords, keywords are set to an empty array
- [x] A new unit test is added to cover this behavior
- [x]  The `gatsby-build` quickstart is updated to no longer have any keywords (via triggering republish)


MichelLosier (approved):

d3caf (approved):
N O I C E

PR #2082, opened 2023-09-27
fix kubernetes.json query syntax

### LINKED ISSUE #2081: NRQL Syntax issue with dashboard preventing import
https://github.com/newrelic/newrelic-quickstarts/blob/1401f84c41daa2d1d84dc455859868d1a44af3b9/dashboards/kubernetes/kubernetes.json#L586

The NRQL here appears to be missing an `AND` and is therfore failing to import correctly.

This pat of the NRQL
`where metricName = 'k8s.pod.createdAt' nodeName is not null facet nodeName`

should possibly read:
`where metricName = 'k8s.pod.createdAt' AND nodeName is not null facet nodeName`

csalvador-nr:
this fix #2081

### DIFF
--- a/dashboards/kubernetes/kubernetes.json
+++ b/dashboards/kubernetes/kubernetes.json
@@ -583,7 +583,7 @@
             "nrqlQueries": [
               {
                 "accountId": 0,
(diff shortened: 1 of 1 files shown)


csalvador-nr:
duplicated #2061

PR #2140, opened 2023-11-03
Update elasticsearch.json

csalvador-nr:
# Summary

Fixed the `SELECT` clause from:
Active Search Latency (ms) 
Get Requests Latency (ms) 
Failed Indexing Operations per Node (Filterable) 
Indexing Latency (ms) 
Index Refresh Latency (ms)

Some of them were dividing UNITS/milliseconds. When the goal is to find the average time of each unit it should be MS/UNITS.  Additionally, `host.hostname` was being used for FACET when those metrics resulted in inconsistent data. Switched to using `entity.name` for correct results.  `capture()` is used to sanitise node names.


mickeyryan42 (approved):

rahulbasu-nr (approved):
Looks good to me.

PR #2162, opened 2023-12-08
chore: fixing issue with biztalk360

jcountsNR:
# Summary

SImple fix for biztalk ID


jcountsNR:
I think there is a problem with the check here. It's saying that the display name is already taken, which not really true. I'm just trying to change the id because I have to reference it in the marketplace.

ERROR: The following errors occurred while validating: biztalks360
- dataSourceMetadata/displayName: `displayName` has already been taken

jcountsNR:
@rahulbasu-nr , I'm not sure if you know the answer to this one. Someone put the id for the datasource here as `third-party-biztalk360`, and I'm trying to fix it into something more usable for the marketplace. Is it possible to change the ID of a data-source once it's been set? And if so, how can I resolve the schema compliance error?

PR #2165, opened 2023-12-14
Update and rename mssql-queryplan.json to mssql-customqueries.json

csalvador-nr:
This change is due to the fact that this quickstart was initially created from an [experimental integration](https://github.com/newrelic-experimental/nri-mssql-experimental/tree/master) that sends information about Query Plans in `Logs`. However, in this dashboard these Query Plans Logs are not queried anywhere. I have found that this is generating a lot of confusion among users who expect that after enabling this integration these widgets will be populated. 
The reality is that this dashboard is powered by the basic MSSQL integration + the custom SQL queries that are executed in [this file](https://github.com/newrelic/nri-mssql/blob/master/mssql-custom-query.yml.sample).
I have changed the name of the dashboard for clarification and added as a side note warning that this file must be used for the dashboard to contain data in its entirety.
image


jbeveland27:
Hey @csalvador-nr - thanks for making a new PR. I'm fine with leaving the quickstart in place, just want to remove the experimental integration since we've archived it and are no longer supporting it. I'm not sure how this quickstart and the associated dashboard came to be (it predates my time on this team), but the integration was specifically around gathering Query Plans as Log data. Since the Query Plans part of this is no longer in play that changes the intent of this quickstart (IMO), and just turns this into more of an opinionated dashboard to go along with the custom queries file from nri-mssql. Not a big deal either way, just wanted to put it out there that this effectively changes the definition of what the quickstart is and does.

A couple of minor notes on the PR changes:
- The quickstart definition was removed here (`quickstarts/mssql-queryplan/config.yaml`). I think you'll want to add that back (with the `Custom MSSQL Queries` name), otherwise this won't show on the public I/O site anymore.
- The dashboard requires the nri-mssql to work, so I believe that should be referenced as the data source in the quickstart definition. However, I'm not aware of a way to indicate to the user that the custom query file needs to be included apart from what you did with the note on the dashboard. Maybe the team managing the quickstarts can give a pointer there.

csalvador-nr:
Thank you for the commen @jbeveland27, I agree that the fact that it is a dashboard that requires extra queries could change the purpose of the quickstarts. However, since the work to create the dashboard and publish it has already been done based on those custom queries, I think we lose nothing by keeping the dashboard and simply putting a notice that it requires running those queries.

I have added the suggested changes, including the definition of the quickstart so that it appears in the I/O.

data-sources/mssql-customqueries/config.yml
```diff
@@ -0,0 +1,9 @@
+id: mssql-customqueries
+displayName: MSSQL On-Host Integration w/ Custom Queries
+description: |
+  Display MSSQL information collected from custom queries
+install:
+  primary:
+    link:
+      url: https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/microsoft-sql/microsoft-sql-server-integration/
+icon: logo.svg
```
jbeveland27, line 9: @csalvador-nr Sorry, this isn't what I meant. What you've added here is a `data source`, and there's an existing one for nri-mssql already. What you need to do is add `quickstarts/msqql-customqueries/config.yml` (and logo file) and add a quickstart definition that references the mssql data source ID [like so](https://github.com/newrelic/newrelic-quickstarts/blob/release/quickstarts/mssql/config.yml#L25-L26).

jbeveland27: For ref, by using this existing data source, clicking `Install` from the public site will allow the user to get to the in-product install experience for MSSQL:
image

csalvador-nr: My bad, I believe commit c8b0bbf fix it.

quickstarts/mssql-customqueries/config.yml
zstix, line 1: Can you rename the directory name to match the `slug` for this quickstart?

Right now the directory is `mssql-customqueries` and it should be `mssql-custom-queries`.

zstix:
Hey @jbeveland27! Other than the small comment I made about the directory name, does this quickstart look good to you? I'm a little unclear on the process for this PR.

jbeveland27:
> Hey @jbeveland27! Other than the small comment I made about the directory name, does this quickstart look good to you? I'm a little unclear on the process for this PR.

Hey @zstix! Long time no see 😄 Sorry for the confusion, this PR came about after I made #2167 to remove the quickstart altogether (it was using an integration my team archived). I left a few comments previously to help @csalvador-nr align the quickstart with the appropriate `microsoft-sql-server` dataSourceId so it still works.

---

In addition to your comment about renaming the directory, there's a few more consistency changes I'd advise for @csalvador-nr:
- rename the dashboard directory from `dashboards/mssql-customqueries` to `dashboards/mssql-custom-queries` to make things consistent
- remove `data-sources/mssql-query-plans` directory altogether (there's still a `logo.svn` file). This data source and all assets should be completely purged since it's been archived.

csalvador-nr:
I changed all names to `mssql-custom-queries` for consistency. Additionally, I updated the screenshots. @jbeveland27 @zstix

rahulbasu-nr (approved):
Looks good to me.

brammerl (approved):
LGTM seems like all the feedback was addressed.

jbeveland27 (approved):
🚢 🇮🇹

csalvador-nr:
not stale

PR #2179, opened 2023-12-28
NR-199374 // Release script

d3caf:
This adds a new `release` script which can be invoked via `npm run release` or `yarn release`. The script will run through a wizard to ensure that all required env vars are set and will target the selected environment for release. It relies on there being a merged PR from `release` -> `main` and requires the PR number.

## Testing
- `npm install` in `utils`
- Have a staging API token handy
- `npm run release`
- Can test against #2181 by providing it as the PR number when prompted. The script only uses the PR to determine which files changed and then runs mutations based on local files, so changes may not actually be accurate unless you've pulled the changes from 2181 locally. 

## TODO
- [x] Test against local NG
- [x] ~Commit documentation~ Docs will live in Confluence and be written as a part of another ticket


utils/package-lock.json
aswanson-nr, line 1: I think this may have been accidentally created, let's remove it!

utils/release.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  failures['dataSources'] = await createValidateDataSources(url, token, dryRun);
+  failures['quickstarts'] = await createValidateQuickstarts(url, token, dryRun);
+
+  if (!dryRun) {
+    failures['dashboardsRequiredDatasources'] =
+      await setDashboardsRequiredDataSources(url, token);
+
+    failures['alertPoliciesRequiredDatasources'] =
+      await setAlertPoliciesRequiredDataSources(url, token);
```
aswanson-nr, line 153: In the pipelines, a single failure in any of these would stop the subsequent ones from running. Do you think that's worth maintaining here, or are we covered with the initial dry run?

aswanson-nr (commented):
LGTM! One straggler file that needs to be removed, but that it should be ready!

utils/release.ts
d3caf: @aswanson-nr Yeah, true. I'm gonna remove this check here since it'll crash and burn anyway.

aswanson-nr:
Let's also disable the failing jobs so that they don't create a lot of noise. We can just comment out the follow lines: https://github.com/newrelic/newrelic-quickstarts/blob/release/.github/workflows/pr-merged.yml#L113-L165

aswanson-nr (approved):

caylahamann (approved):
The only thing is maybe we should mention / document what the github token is used for and if there is any scenario in which you might need it?? just because i got tripped up on that one

PR #2184, opened 2024-01-04
fix: Target `release` and update checks

aswanson-nr:
# Summary

Now that pull requests are going into the `release` branch, we need to generate quickstart ids at that time instead when they are merged into the `main` branch. This ensures that the `release` branch stays ahead or even with `main`. 
Also updates the required checks to be accurate.

### DIFF
--- a/.github/workflows/pr-merged.yml
+++ b/.github/workflows/pr-merged.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           token: ${{ secrets.OPENSOURCE_BOT_TOKEN }}
-          ref: "main"
+          ref: "release"
           fetch-depth: 0
 
       - name: Setup workspace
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-merged.yml
```diff
@@ -80,18 +80,19 @@ jobs:
...
               repo: context.repo.repo,
-              branch: 'main',
+              branch: 'release',
               required_status_checks: {
                 strict: true,
                 contexts: [
                     'Validation / Image count and extension compliance',
                     'Validation / Ensure icons exist',
                     'Validation / Install plan ids exist',
-                    'Validation / Validate data source ids',
```
josephgregoryii, line 90: thank you

josephgregoryii (approved):

josephgregoryii (approved):

PR #2208, opened 2024-01-11
Cassandra queries back to events

obrahc:
# Summary

Changing metrics query back to Sample events for the Cassandra Quickstart.

## Pre merge checklist


zstix (approved):
These are some pretty big query changes. I'm not able to review all of these so I'm trusting the author and linter a lot here.

PR #2231, opened 2024-01-11
Some fixes to the release script

aswanson-nr:
# Summary
* Reducing the number of requests that run at a time to avoid race conditions in the API. 
* Filter out test/mock files
* Update alert validation to look for the `TIMESERIES` keyword not just the word `timeseries`

### DIFF
--- a/utils/create_validate_pr_quickstarts.ts
+++ b/utils/create_validate_pr_quickstarts.ts
@@ -3,7 +3,7 @@ import {
   filterOutTestFiles,
   isNotRemoved,
 } from './lib/github-api-helpers';
-import { translateMutationErrors, chunk } from './lib/nr-graphql-helpers';
+import { translateMutationErrors } from './lib/nr-graphql-helpers';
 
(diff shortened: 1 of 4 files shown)


utils/create_validate_pr_quickstarts.ts
```diff
@@ -125,13 +127,12 @@ export const createValidateQuickstarts = async (
   const quickstartErrors: string[] = [];
 
   logger.info(`Submitting ${quickstarts.length} quickstarts...`);
-  for (const c of chunk(quickstarts, 5)) {
```
josephgregoryii, line 128: Since we're not using `chunk` here anymore, we may be able to remove the import at the top of the file.

josephgregoryii (approved):

josephgregoryii (changes requested):
Undoing my approval - seeing some failing tests?

utils/create_validate_pr_quickstarts.ts
aswanson-nr: Done!

josephgregoryii (approved):

PR #2244, opened 2024-01-17
Remove adding label to open pr

josephgregoryii:
# Summary

The original AC for this ticket was to have our workflow use node 16.
Unfortunately, the action script we used to get modified/added files in PRs was abandonded and no longer had any maintanance on it. That action was using node 12, which is deprecated.

Since we would need to add a SR to add in a _different_ 3rd party action script, decided that having automatic tagging wasn't worth that potentially long tail end to get it reviewed.

This PR removes that action and the file along with it


brammerl (approved):
While this is all fine and dandy I'm curious if we do eventually want to attempt to bring tagging back?? 

Should we create a ticket to explore how we would do that/reviving tagging?

PR #2299, opened 2024-02-24
Updating 4 partner data sources to point to their Install configs 

JuliaNocera:
# Summary

This PR updates: Cribl, Delphix, Gigamon App Insights and Netlify Builds data sources to use the install-data-source install experience.

_The owners of this repo are not experts in the subject matter of the quickstarts. We review for the quickstart to be functional and for security risks. If you are seeking feedback on the content of the quickstart, please seek out a subject matter expert. If you are not an internal NR contributor, we can do our best to connect you with a content reviewer._


mickeyryan42 (commented):
This PR lgtm, however in the install-data-source nerdlet we have `gigamon` being mapped to the install and not `gigamon-appinsights`.  
We would want to update in one or the other location, probably on the mapping side.  So if thats the route we take i dont think we need any changes on this PR, but we'd want to make a PR to marketplace first to ensure the mapping is correct.

mickeyryan42 (approved):

PR #2312, opened 2024-03-06
LLM datasource updates

d3caf:
We've created a new config for Langchain and Bedrock to use so updating their datasources to point to that.

We've additionally created a new datasource for the Azure OpenAI quickstart to use.


natecanfield822 (approved):
Excited to see this coming together!

brammerl (approved):

brammerl (approved):

PR #2390, opened 2024-04-30
Update DupMetrics.yml

relic-js:
# Summary

As written, the alert condition in this YAML is very computationally expensive, because it matches data points from all metrics that report `job`, `instance`, and `cluster_name` attributes. I've added an additional `WHERE` clause that scopes the query down to just Prometheus metrics, since Prometheus metrics are all that we care about in this particular context.


pkudikyala (approved):
LGTM!

sjyothi54 (approved):
LGTM

PR #2409, opened 2024-05-09
Release 2024-05-09

zstix:


zstix:
We can safely ignore these warnings about `guid` and `entityGuid` since the offending NRQL statements are only using these fields in a `unique()` function (not checking against a _specific_ entity).

seemantk (approved):

PR #2464, opened 2024-06-28
Add missing datasources [NR-286163]

d3caf:
We have a number of quickstarts that have no associated datasource. This PR adds docs-based datasources and associates them with the proper quickstart.


caylahamann (commented):
do we need to add the `nr1_addData` keyword or is that no longer necessary?

d3caf:
> do we need to add the `nr1_addData` keyword or is that no longer necessary?

@caylahamann I didn't because these are new data sources solely to back up some quickstarts. Since these weren't in product before, I figured we wouldn't add them.

caylahamann (approved):

mickeyryan42 (approved):

PR #2474, opened 2024-07-09
feat: Remove keywords if they're not present

d3caf:
If keywords aren't provided, then submit an empty array so they get cleared in the database.


zstix (dismissed):

mickeyryan42 (approved):

utils/lib/DataSource.ts
```diff
@@ -97,7 +97,7 @@ class DataSource extends Component<DataSourceConfig, string> {
       icon: this._getIconUrl(),
       install: this._parseInstall(),
       categoryTerms: categoryTerms && categoryTerms.map((t) => t.trim()),
-      keywords: keywords && keywords.map((k) => k.trim()),
+      keywords: keywords ? keywords.map((k) => k.trim()) : [],
```
aswanson-nr, line 100: I know the ticket doesn't call this out, but can we also do this for quickstarts?
Also thinking `categoryTerms` should get the same treatment...

d3caf: @aswanson-nr it looks like quickstarts already have this behavior https://github.com/newrelic/newrelic-quickstarts/blob/269e82d447298281c9af044db1f64f5a738ecb87/utils/lib/Quickstart.ts#L176

aswanson-nr (approved):

PR #2495, opened 2024-07-22
Lucia/migrate wikis

brammerl:
https://new-relic.atlassian.net/browse/NR-291537


zstix:
Perhaps we should _not_ have these links in the repository, since this is a public-facing repository. I'm fine not having this information linked, personally.

PR #2538, opened 2024-08-27
Introduce JSON schemas for quickstart artifacts

zstix:


mickeyryan42:
> How do we want to organize the schema directory?
This is what I imagined as well, unless we have a reason to go a different way this seems to be easiest to maintain. 

> How do we want to reference other schemas?

I may be misunderstanding this, but is this saying if we use $refs then those entire sub schemas wont be validated from the top level?  
`$ref property, the object is considered a reference, not a schema. Therefore, any other properties you put in that object will not be treated as JSON Schema keywords and will be ignored by the validator.`

> How much detail do we want to provide?

I agree with being mostly specific about details, it seems like the point of defining an artifact we can validate that we want to be able to fail on unexpected or missing fields.  That said we can probably be more loose on certain things that we won't break on should they be.  I could see additionalProperties being ok since we would just ignore additional fields on the backend.

> What version (and features) of JSON schema do we want to embrace?

I'm not entirely sure about this and would be interested to see side by side differences.  I do see this though, which is interesting: `Optionally, you can add draft-07 meta-schema, to use both draft-07 and draft-2019-09 schemas in one Ajv instance:`

> Are we alright using extra package(s) for improved validation?

Ajv has a number of extensions/plugins they list in their docs https://ajv.js.org/packages/#extending-ajv
Imo especially since we aren't worrying about a bundle size or package vulnerabilities in a production environment we should go for whatever we find useful as long as its maintained and credible.  One other consideration is whether we would need the same type of extension on the catalog service side and whether its available in Elixir, or we can accomplish the same task without it easily enough.

zstix:
Decisions from a zoom:

- Maintain a list of "core" datasource IDs
- When we go to build the artifact:
  - Also build a schema that has an `enum` of _all_ datasourceIds
  - Add in the core datasource IDs to the enum
- Publish _both_ the artifact and the (lightly) generated schema

The schema is mostly static, the only thing we actually "generate" is the list of datasource IDs.

utils/schema/core-datasource-ids.json
d3caf, line 1: So will this need to be manually kept in sync with catalog service's core datasources? If so, I'd worry about drift and introducing more manual steps to keep track of, although I suppose core datasources don't change often...

utils/schema/artifact.json
```diff
@@ -0,0 +1,194 @@
...
+          "type": "array", "items": { "type": "string" }
+        },
+        "icon": { "type": "string" }
+      },
+      "required": ["id", "displayName", "install"],
+      "additionalProperties": false
+    },
+
+    "installDirective": {
+      "enum", [
```
d3caf, line 115:
```suggestion
      "enum": [
```
Should this be changed?

zstix: Yes?

utils/schema/core-datasource-ids.json
zstix: This was the path we agreed upon in the campfire review on Wednesday. I'm not enamored with the solution either, but I'm not sure of an alternative approach. If you have any suggestions, please let me know!

The good thing is that, if we forget to keep this in sync, any quickstart deployment that makes use of the new datasource will fail...so there's not really a chance of us missing this and it making it to production.

d3caf: Ah, with the early failure, I'm good with this solution.

docs/README.md
```diff
@@ -1,3 +1,10 @@
 # Documentation
 
 Each file in this directory (minus this readme) is generated from New Relic NerdGraph GraphQL schema.
+
+## Validation
+
+Every quickstart, datasource, alert, and dashboard is bundled into a single artifact and validated against our [schema](../utils/schema/artifact.json) for correctness.
+
+> [!IMPORTANT]
+> Please ensure that _all_ feature flags are removed from a datasource before it is referenced in a quickstart. Otherwise, the datasource will not be properly associated.
```
d3caf, line 10: [nit] Since only core datasources can use FF's, maybe we should consider clarifying this a bit as it could cause confusion for community datasources. I also wonder if we want to be a bit more explicit about the issues caused by not fully enabling a FF.

> For quickstarts referencing a `core` datasource, ensure that _all_ datasource feature flags are removed 100% enabled. Failure to do so will result in the quickstart failing to load properly in the UI.

Not a blocker, just thought maybe it could be a bit more clear.

d3caf (dismissed):

zstix:
Some additional questions, and my attempts to answer them:

> We seem to be using the capitalization of `datasourceIds` in the schema, but in the actual quickstart definitions, it's `dataSourceIds` (this always feels like a painpoint)

This was a conscious choice, but made purely out of my own personal preference. I suppose we probably _should_ update this to be `dataSourceIds` since that's how it's represented in NerdGraph 😩 

> We require a datasource ID on a quickstart, but it seems like we currently have quickstarts that aren't associated to a datasource ID so the artifact is failing validation. I think it's valid for a quickstart to not have a datasource ID in the case that it's a docs-only quickstart?

`datasourceIds` are not one of the [required fields](https://github.com/newrelic/newrelic-quickstarts/pull/2538/files#diff-483226189838dd862a85d2973c687dfdac4a0354a604074f70e351326f4b9249R73-R81) for the quickstart schema.

utils/schema/core-datasource-ids.json
```diff
@@ -0,0 +1,140 @@
+[
```
josephgregoryii, line 1: Out of curiosity, are all these core data sources visibile to all customers?

Unsure if there was already discussion about only allowing core data sources that have their `visibility` flags set to `deployed` / 100% traffic. 

Not a blocking comment, and since we call it out in the docs to ensure that core data sources are fully deployed and visible, I see this more on the contributor to do their due-diligence.

josephgregoryii (commented):
Not a leading question, just curious if we had anything set for this question.

d3caf:
@zstix 
> datasourceIds are not one of the [required fields](https://github.com/newrelic/newrelic-quickstarts/pull/2538/files#diff-483226189838dd862a85d2973c687dfdac4a0354a604074f70e351326f4b9249R73-R81) for the quickstart schema.

Huh. I'm getting the following error in the validation.
```
Error: schema is invalid: data/properties/datasourceIds/enum must NOT have fewer than 1 items
    at Ajv.validateSchema (/Users/aanguiano/c/newrelic-quickstarts/utils/node_modules/ajv/lib/core.ts:519:18)
    at Ajv._addSchema (/Users/aanguiano/c/newrelic-quickstarts/utils/node_modules/ajv/lib/core.ts:721:30)
    at Ajv.compile (/Users/aanguiano/c/newrelic-quickstarts/utils/node_modules/ajv/lib/core.ts:384:22)
    at Ajv.validate (/Users/aanguiano/c/newrelic-quickstarts/utils/node_modules/ajv/lib/core.ts:361:16)
    at main (/Users/aanguiano/c/newrelic-quickstarts/utils/build-validate-quickstart-artifact.ts:31:7)
    at Object.<anonymous> (/Users/aanguiano/c/newrelic-quickstarts/utils/build-validate-quickstart-artifact.ts:37:3)
    at Module._compile (node:internal/modules/cjs/loader:1233:14)
    at Module.m._compile (/Users/aanguiano/.local/share/mise/installs/node/20.5.1/lib/node_modules/ts-node/src/index.ts:1618:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1287:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/aanguiano/.local/share/mise/installs/node/20.5.1/lib/node_modules/ts-node/src/index.ts:1621:12)
```

I thought that was coming from the quickstart definition but I guess it's not. The top-level `datasourceIds` field has an array of datasource ID strings in it so I'm confused what the issue is.

zstix:
@d3caf Enums require at least one entry. We plan on populating that field at build time (the script included in this PR) so that should always have items in it.

utils/schema/core-datasource-ids.json
zstix: Memory is a bit hazy, but I think we ran some scripts to get all the IDs and we _might_ have factored in the existence of a feature flag.

That said, short of clicking into each feature flag, I'm not sure how we would be able to verify the traffic %. Totally open to suggestions on how to improve tho!

utils/schema/artifact.json
```diff
@@ -0,0 +1,194 @@
...
+    "quickstarts",
+    "datasources",
+    "alerts",
+    "dashboards",
+    "datasourceIds"
+  ],
+  "additionalProperties": false,
+
+  "$definitions": {
+    "quickstart": {
```
d3caf, line 34: We currently appear to have some quickstarts that still have the `installPlans` field so schema validation will fail. I'm not exactly sure how we want to handle that.

d3caf (commented):
Still have some open questions so switching to a comment for the review

zstix:
### Next steps
- [ ] Cherry-pick commit from other branch for schema
- [ ] Ensure consistent `dataSource` naming (capital `S`)

utils/schema/core-datasource-ids.json
josephgregoryii: Top of my head i cant see think of a way without creating some type of script which _could_ be over-engineer to figure out this "simple" edge case.

I think it could be something to look into at another point, but def not worth the scope of this ticket. Maybe once we get some sort of validaiton on the elixir side we could add it, but not sure if we could necessarily catch it at PR time.

utils/schema/add-datasource-ids.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import type { DataSourceConfig } from '../types/DataSourceConfig';
+
+import { getAllDataSourceFiles } from '../lib/DataSource';
+
+const ARTIFACT_SCHEMA_PATH = './artifact.json';
+const CORE_DATASOURCE_IDS_PATH = './core-datasource-ids.json';
+const DATASOURCE_BASE_PATH = '../../';
+
+// TODO: infer the type from the schema?
```
josephgregoryii, line 12: Is this TODO still valid, or something we'll address at a later point?

utils/schema/add-datasource-ids.ts
```diff
@@ -0,0 +1,84 @@
+import * as fs from 'fs';
+import * as yaml from 'js-yaml';
+
+import type { DataSourceConfig } from '../types/DataSourceConfig';
+
+import { getAllDataSourceFiles } from '../lib/DataSource';
+
+const ARTIFACT_SCHEMA_PATH = './artifact.json';
+const CORE_DATASOURCE_IDS_PATH = './core-datasource-ids.json';
+const DATASOURCE_BASE_PATH = '../../';
```
josephgregoryii, line 10: Not sure if its been a while in this repo, but i wasn't able to get this script to run right away.

I'm getting an error running the script using 

```bash
$ cd utils/
$ yarn run ts-node schema/add-datasource-ids.ts 
```

Error:
```bash
yarn run v1.22.19
$ /Users/jgregory/Developer/newrelic-external/newrelic-quickstarts/utils/node_modules/.bin/ts-node schema/add-datasource-ids.ts
[*] Found 0 community dataSources
Error: ENOENT: no such file or directory, open './core-datasource-ids.json'
    at Object.openSync (node:fs:585:3)
    at Object.readFileSync (node:fs:453:35)
    at getCoreDataSourceIds (/Users/jgregory/Developer/newrelic-external/newrelic-quickstarts/utils/schema/add-datasource-ids.ts:38:8)
    at main (/Users/jgregory/Developer/newrelic-external/newrelic-quickstarts/utils/schema/add-datasource-ids.ts:69:19)
    at Object.<anonymous> (/Users/jgregory/Developer/newrelic-external/newrelic-quickstarts/utils/schema/add-datasource-ids.ts:83:3)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Module.m._compile (/Users/jgregory/Developer/newrelic-external/newrelic-quickstarts/utils/node_modules/ts-node/src/index.ts:1618:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/jgregory/Developer/newrelic-external/newrelic-quickstarts/utils/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:981:32) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: './core-datasource-ids.json'
}
```

I was able to run the script by updating these lines to utilize `path`

```ts
const ARTIFACT_SCHEMA_PATH = path.resolve(__dirname,'artifact.json');
const CORE_DATASOURCE_IDS_PATH = path.resolve(__dirname,'core-datasource-ids.json');
const DATASOURCE_BASE_PATH = path.resolve(__dirname, '..', '..');
```

josephgregoryii (changes requested):
Code looks good, changing the paths using `path` (see comment below) I saw the reflected changes in the `artifact.json` file which added core and community data sources to the artifact.

One comment about the `TODO`. Not blocking the PR for it but if we don't need the `TODO` anymore we can consider dropping that comment. 

The requested changes has to do with not being able to run `add-datasource-ids.ts` in the terminal, which i would expect it to break when we add it to a workflow / action.

utils/schema/add-datasource-ids.ts
zstix: Oh good point, I suppose that is up for discussion. Since Typescript is _structurally_ typed, and because we only care about the type defined here, then I'm not sure if it's worth generating the type for the entire schema.

What do you think? If you agree, I can remove the `TODO`

utils/schema/add-datasource-ids.ts
zstix: I believe the intention was to run `ts-node` from the same directory as the script (just verified, this still works). If we would prefer, I can update the script to work from the parent directory instead.

If it would be easier, I can also add this as a npm script.

utils/schema/add-datasource-ids.ts
josephgregoryii: I care more about the artifact being structurally correct, so im fine to leave it as is and remove the TODO :)

utils/schema/add-datasource-ids.ts
josephgregoryii: Ah i see, yeah totally works by just running `ts-node add-datasource-ids.ts` in the same location as the script. It just doesn't work with `yarn run ts-node` (which makes total sense), or in a parent directory. 

My original concern with not using `path` _was_ that we use ubuntu` to run our workflows. I thought it had a similar file system traversing as windows (which having `../` would break), but TIL its similar to mac. Also, doesn't seem like we're going to use this specific script in a workflow.

I could've verified at the time that the file structure is similar, so its not _really_ a concern anymore.

Having an npm script could be nice if we end up using this specific script in a workflow, but It doesn't seem like we are so we can skip that. Thanks for the clarification!

josephgregoryii (dismissed):
Verified the desired intention with running the script. Looks good!

utils/schema/add-datasource-ids.ts
zstix: I've added an npm script to make this a little easier.

utils/schema/core-datasource-ids.json
mickeyryan42: Flags weren't taken into account so this is just all core data sources.  I imagine its fine to show the IDs, flags being more for hiding the experience or visibility.  Feature flags also generally contain the data source id (in most cases) and are visible in the network.  Since this is schema related and not building a list for the client, don't we need all core data source IDs to be present?

mickeyryan42 (approved):

PR #2540, opened 2024-08-29
Artifact creation and validation

d3caf:
This is based on #2538 and adds a script to create the release artifact and validate it.

## Todo
- [x] Validate artifact with JSON schema
- [x] Verify that artifact is properly created
- [x] ~Save artifact to filesystem (or in a way that can be published easily by the next step in the process)~
- [x] ~Set up GH action to run script~

### DIFF
--- a/utils/__tests__/build-validate-quickstart-artifact.test.js
+++ b/utils/__tests__/build-validate-quickstart-artifact.test.js
@@ -1,0 +1,163 @@
+import * as fs from 'fs';
+import Quickstart from '../lib/Quickstart';
+import DataSource from '../lib/DataSource';
+import Alert from '../lib/Alert';
+import Dashboard from '../lib/Dashboard';
+
+import {
+  getArtifactComponents,
+  getDataSourceIds,
+  validateArtifact,
+} from '../build-validate-quickstart-artifact';
+
+jest.mock('../lib/Quickstart');
+jest.mock('../lib/DataSource');
+jest.mock('../lib/Alert');
+jest.mock('../lib/Dashboard');
+jest.mock('fs');
+
+describe('built-validate-quickstart-artifact', () => {
+  beforeEach(() => {
+    // disable normal logging used as feedback while script runs
+    jest.spyOn(console, 'log').mockImplementation(() => {});
+  });
+
+  describe('build', () => {
+    it('should find all of the components', () => {
+      Quickstart.getAll = jest
+        .fn()
+        .mockReturnValueOnce([
+          { config: 'test-quickstart-1' },
+          { config: 'test-quickstart-2' },
+        ]);
+
+      DataSource.getAll = jest
+        .fn()
+        .mockReturnValueOnce([{ config: 'test-dataSource-1' }]);
+
+      Alert.getAll = jest.fn().mockReturnValueOnce([]);
+      Dashboard.getAll = jest.fn().mockReturnValueOnce([]);
+
+      const actual = getArtifactComponents();
+
+      expect(actual.quickstarts).toHaveLength(2);
+      expect(actual.quickstarts[0]).toEqual('test-quickstart-1');
+      expect(actual.quickstarts[1]).toEqual('test-quickstart-2');
+      expect(actual.dataSources).toHaveLength(1);
+      expect(actual.dataSources[0]).toEqual('test-dataSource-1');
+      expect(actual.alerts).toHaveLength(0);
+      expect(actual.dashboards).toHaveLength(0);
+    });
+
+    it('should produce a complete list of dataSource IDs', () => {
+      Quickstart.getAll = jest.fn().mockReturnValueOnce([]);
+      Alert.getAll = jest.fn().mockReturnValueOnce([]);
+      Dashboard.getAll = jest.fn().mockReturnValueOnce([]);
+      DataSource.getAll = jest
+        .fn()
+        .mockReturnValueOnce([
+          { config: { id: 'community-1' } },
+          { config: { id: 'community-2' } },
+          { config: { id: 'community-3' } },
+        ]);
+
+      const { dataSources } = getArtifactComponents();
+      fs.readFileSync.mockReturnValueOnce(JSON.stringify(['core-1', 'core-2']));
+
+      const actual = getDataSourceIds('dummy-file.json', dataSources);
+
+      expect(actual).toHaveLength(5);
+      expect(actual).toContain('community-1');
+      expect(actual).toContain('community-2');
+      expect(actual).toContain('community-3');
+      expect(actual).toContain('core-1');
+      expect(actual).toContain('core-2');
+    });
+  });
+
+  describe('validate', () => {
+    const TEST_SCHEMA = {
+      type: 'object',
+      properties: {
+        quickstarts: { type: 'array' },
+        alerts: { type: 'array' },
+        dashboards: { type: 'array' },
+        dataSources: {
+          type: 'array',
+          items: {
+            type: 'object',
+            properties: {
+              id: { type: 'string' },
+              title: { type: 'string' },
+            },
+          },
+        },
+        getDataSourceIds: { type: 'array', items: { type: 'string' } },
+      },
+    };
+
+    it('should correctly build and validate valid artifacts with no errors', () => {
+      Quickstart.getAll = jest.fn().mockReturnValueOnce([]);
+      Alert.getAll = jest.fn().mockReturnValueOnce([]);
+      Dashboard.getAll = jest.fn().mockReturnValueOnce([]);
+      DataSource.getAll = jest
+        .fn()
+        .mockReturnValueOnce([
+          { config: { id: 'community-1', title: 'DataSource 1' } },
+          { config: { id: 'community-2', title: 'DataSource 2' } },
+          { config: { id: 'community-3', title: 'DataSource 3' } },
+        ]);
+
+      const components = getArtifactComponents();
+
+      fs.readFileSync.mockReturnValueOnce(JSON.stringify(['core-1', 'core-2']));
+      const dataSourceIds = getDataSourceIds(
+        'dummy-file.json',
+        components.dataSources
+      );
+
+      const artifact = { ...components, dataSourceIds };
+
+      const actual = validateArtifact(TEST_SCHEMA, artifact);
+
+      expect(actual).toHaveLength(0);
+    });
+
+    it('should correctly return errors for an invalid artifact', () => {
+      Quickstart.getAll = jest.fn().mockReturnValueOnce([]);
+      Alert.getAll = jest.fn().mockReturnValueOnce([]);
+      Dashboard.getAll = jest.fn().mockReturnValueOnce([]);
+      DataSource.getAll = jest
+        .fn()
+        .mockReturnValueOnce([
+          { config: { id: 'community-1', title: 'DataSource 1' } },
+          { config: { id: false, title: 'DataSource 2' } },
+          { config: { id: 'community-3', title: 3 } },
+        ]);
+
+      const components = getArtifactComponents();
+
+      fs.readFileSync.mockReturnValueOnce(JSON.stringify(['core-1', 'core-2']));
+      const dataSourceIds = getDataSourceIds(
+        'dummy-file.json',
+        components.dataSources
+      );
+
+      const artifact = { ...components, dataSourceIds };
+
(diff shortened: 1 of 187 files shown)


utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (ajv.errors?.length) {
+    console.error('*** Validation failed. See errors below. ***');
+    console.error('--------------------------------------------');
+    parseErrors(ajv.errors, artifact);
+
+    process.exit(1);
+  }
+}
+
+const parseErrors = (errors: ErrorObject[], artifact: Record<string, any>) => {
```
d3caf, line 42: Down to refactor this if folks feel like it's necessary.

d3caf:
## Status
I have the validations running properly and outputting the errors to the console. Next steps would be:

- [ ] Fix existing quickstarts that are failing validations (they're actually invalid, we just never validated them super closely)
- [ ] Code cleanup/refactor
- [ ] Evaluate whether we want to use `json-schema-to-ts` to generate the TS interface for the artifact (may be unnecessary)
- [ ] Save the artifact in a way that GH actions can use it
- [ ] Set up GH action

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const invalidItem = artifact[artifactItemPath[0]][artifactItemPath[1]];
+
+    console.error(`Error #${idx + 1}:`, e);
+    console.error('                         ');
+    console.error('Received value:', badValue);
+    console.error('                         ');
+    // All of our properties in the artifact are arrays so we can make some
+    // assumptions to grab the invalid item from the artifact
+    if (invalidItem !== badValue) {
```
sarahkitten, line 108: The logic here / these variable names / the relation to the comment above are a bit confusing to me!  Maybe the comment could be clarified and/or the names could be changed (maybe `invalidItemPath`, `invalidItem`, and `invalidValue`?)

utils/lib/Alert.ts
```diff
@@ -297,6 +297,10 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
 
     return result;
   }
+
+  static getAll() {
```
sarahkitten, line 301: This doesn't seem to be working since it's coming back with 0 alerts

utils/build-validate-quickstart-artifact.ts
d3caf: Yeah looks like this may have gotten stuck in the wrong place during a refactor

utils/lib/DataSource.ts
```diff
@@ -41,7 +41,9 @@ class DataSource extends Component<DataSourceConfig, string> {
 
     // find the matching data source ID from the config content
     const dataSource = allDataSources.find((i) => i.content?.id === id);
-
+    if (!dataSource) {
+      console.log('DATA SOURCE ID: ', id);
```
mickeyryan42, line 45: debug stuff

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import Quickstart from "./lib/Quickstart";
+import DataSource from "./lib/DataSource";
+import Alert from "./lib/Alert";
+import Dashboard, { DashboardConfig } from "./lib/Dashboard";
+import Ajv, { type ErrorObject } from 'ajv';
+import { QuickstartConfig, QuickstartConfigAlert } from './types/QuickstartConfig';
+import { DataSourceConfig } from './types/DataSourceConfig';
+
+type ArtifactSchema = any;
```
mickeyryan42, line 12: Do we want to type this out? I think id be fine either way since its just defining the type of the schema.

d3caf: I was just gonna leave it as `any` for now and we can determine if we wanna type it further depending on whether we need to use the type anywhere else. I didn't see much harm in just leaving this definition in for now though.

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+  const errors = validateArtifact(schema, artifact);
+
+  if (errors.length) {
+    console.error('*** Validation failed. See errors below. ***');
+    console.error('--------------------------------------------');
+
+    parseErrors(errors, artifact);
+
+    process.exit(1);
+  }
```
mickeyryan42, line 84: It may be nice to just output something about successful validation on exit

utils/lib/Alert.ts
```diff
@@ -297,6 +297,16 @@ class Alert extends Component<QuickstartConfigAlert[], QuickstartAlertInput[]> {
...
+    const alertPaths = glob.sync(path.join(__dirname, '..', '..', 'alert-policies', '**', '*.+(yml|yaml)'));
+    return alertPaths.map(alertPath => {
+      // The identifier for alerts is the folder and the file name
+      // e.g. `node-js/HighCpuUtilization.yml`
+      const identifier = path.join(...alertPath.split('/').slice(-2, -1));
```
zstix, line 306: I know it's not _likely_, but I always get a little anxious seeing `slice` (or `pop`) on arrays since those can technically return `[]` (and `undefined` for `pop`).

Typescript doesn't seem to be angry at this here, but I know that we're using a `!` in the Quickstart file to get around this. We might want to consider some additional error checking (or tests) to ensure we _for sure_ have an identifier.

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const parseErrors = (errors: ErrorObject[], artifact: Record<string, any>) => {
+  return errors.forEach((e, idx) => {
+    const artifactItemPath = e.instancePath.split('/').filter(Boolean).map(segment => {
+      // If the segment is a numerical index, convert it to an int and return
+      // it, otherwise just return the segment string.
+      return parseInt(segment, 10) || segment;
```
zstix, line 92: I think there's a potential bug here:

It looks like we're using `parseInt(segment, 10)` to verify whether or not segment is a number. I assume the thinking is that `parseInt("foobar", 10)` would return `NaN` which is falsy...and that would then cause the `|| segment` to kick in. Unfortunately, `parseInt("0", 10)` will _also_ be falsey because `0` is considered false.

In other words: this code ignores numerical indexes if they're `0`.

I wonder if we can come to this answer for a different direction. I'll try out some options. In the meantime, we could always do something like the following, which avoids this issue:

```ts
const numSegment = parseInt(segment, 10);
return !Number.isNaN(numSegment) ? numSegment : segment;
```

utils/build-validate-quickstart-artifact.ts
zstix: Nitpick: could we use `unknown` over `any`? It's slightly less permissive and I think it better signals intent here.

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    // Reduce over the segments to find the bad value in the artifact
+    const badValue = artifactItemPath.reduce((acc, segment) => {
+      return acc[segment];
+    }, artifact);
```
zstix, line 98: It took me a while to understand what we're doing here...it _seems_ like we have a path (`artifactItemPath`) and we're using that as a way to get a nested value inside a bigger object (`artifact`).

It might be easier for us to just pull in [`_.get`](https://lodash.com/docs/4.17.15#get) from lodash and remove this code block (and the one above). This is what we do with many of our internal repositories that solve similar problems.

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    // All of our properties in the artifact are arrays so we can make some
+    // assumptions to grab the invalid item from the artifact
+    const invalidItem = artifact[artifactItemPath[0]][artifactItemPath[1]];
```
zstix, line 102: I think it would be worthwhile to create some tests for these assumptions (I also think tests were part of the ticket's acceptance criteria).

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    console.error(`Error #${idx + 1}:`, e);
+    console.error('                         ');
```
zstix, line 105: I think `console.anything` will print an entire line. If you're trying to create a new line, you could simply do this with `console.error('');`.

Perhaps we should break out error reporting into it's own function. That might make testing easier too (we can mock the function and use `calledWith`-type utilities to assert we have the correct error).

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+    console.error('Received value:', badValue);
+
+    console.error('                         ');
```
zstix, line 108: Same comment as above.

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+    console.error('--------------------------------------------');
+
+    parseErrors(errors, artifact);
+
+    process.exit(1);
+  }
+}
+
+const parseErrors = (errors: ErrorObject[], artifact: Record<string, any>) => {
+  return errors.forEach((e, idx) => {
```
zstix, line 88: Nitpick: `forEach` returns `void`, so the `return` doesn't really do much for us here.

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,121 @@
...
+    console.error(`Error #${idx + 1}:`, e);
+    console.error('                         ');
+    console.error('Received value:', badValue);
+
+    console.error('                         ');
+    if (invalidItem !== badValue) {
+      console.error('Invalid item:', invalidItem);
+    }
+
+    if (idx + 1 !== errors.length) {
```
zstix, line 113: Maybe it's just how my brain works, but it took me a few moments to understand that we're simply trying to add a "line" unless we're on the last item in the list. I think it was the use of `!==` over `<`. To me "less than the max" is easier to grok than "plus one is equal to the max".

```ts
if (idx < errors.length - 1) {
```

Logic aside, if we're trying to separate a bunch of output by a dividing line, then maybe we should `map` over the errors to get an array of _output_ strings and then `.join` the array with the desired divider. That would prevent us from needing to mess around with indexes.

zstix (changes requested):
I found at least one (potential) bug and a few places where we might want to refactor. Additionally, we don't have any tests for this work (something called out in the acceptance criteria).

I'll probably jump in to make these adjustments since this is blocking other work.

utils/lib/Alert.ts
zstix: Note: I wouldn't block the PR on this!

utils/__tests__/build-validate-quickstart-artifact.test.js
```diff
@@ -0,0 +1,163 @@
...
+        .mockReturnValueOnce([
+          { config: 'test-quickstart-1' },
+          { config: 'test-quickstart-2' },
+        ]);
+
+      DataSource.getAll = jest
+        .fn()
+        .mockReturnValueOnce([{ config: 'test-dataSource-1' }]);
+
+      Alert.getAll = jest.fn().mockReturnValueOnce([]);
```
sarahkitten, line 38: we might want to test each of these with non-empty values here since we had problems with the getAll returning nothing before!

utils/build-validate-quickstart-artifact.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const getInvalidItems = (errors: ErrorObject[], artifact: ArtifactSchema): InvalidItem[] => {
+  return errors.map((error) => {
+    // Get the path to the invalid value from the error `instancePath`.
+    // NOTE: we're using `slice(1)` here to remove the leading `/` in the path.
+    const invalidValuePath = error.instancePath.split('/').slice(1);
+
+    const value = get(artifact, invalidValuePath);
+
+    // Get the specific "component" (e.g. the alert or dashboard) that contains
```
sarahkitten, line 101: I appreciate this refactoring and comments!

sarahkitten (approved):
This is easy to read and passes when run with the current config changes PR.  Small suggestion for the tests but not blocking.

mickeyryan42 (approved):

PR #2548, opened 2024-09-10
feat: Artifact validation PR workflow

d3caf:
This adds a new workflow that runs on PRs in this repo. This workflow validates all the quickstarts, datasources, alerts, and dashboards in this repository and prepares to create an artifact to be released (this portion will be done in an upcoming ticket).


.github/workflows/validate_pr_artifact.yml
```diff
@@ -0,0 +1,16 @@
+name: Validate PR Artifact
+
+on: 
+  - pull_request
+  - workflow_call
```
mickeyryan42, line 5: I'm the one that just mentioned adding this, but I'm realizing we were going to do this when using a reusable workflow.  

I don't think we need it now that we've moved to using an action, since we'll just be reusing that right?  Thats my mistake.

mickeyryan42 (approved):

PR #2549, opened 2024-09-10
feat: Validation and release on main push

d3caf:


.github/workflows/release-artifact.yml
```diff
@@ -0,0 +1,52 @@
...
+name: Release artifact
+on:
+  workflow_call:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+
+# FIXME: We may need a different action for `Create new release tag` because
```
d3caf, line 14: Calling this out to be addressed next

d3caf:
I think this is nearly ready to merge. We'll need to do some PR's into main to fully test the workflow. Currently, it can be tested by tweaking some of the triggers/if's. I think the only thing remaining to do is determine whether we're ok with the semver bump being dynamically determined by semantic commit messages (as [noted above](https://github.com/newrelic/newrelic-quickstarts/pull/2549#discussion_r1752677777))

.github/workflows/release-artifact.yml
d3caf: This may be better addressed in the release workflow ticket

utils/build-validate-quickstart-artifact.ts
```diff
@@ -88,6 +89,24 @@ const main = () => {
...
+  try {
+    fs.mkdirSync('./build', { recursive: true });
+
+    // Dump the artifact to artifact.json
+    fs.writeFileSync('./build/artifact.json', JSON.stringify(artifact));
+    // Copy the schema to schema.json
+    fs.copyFileSync('./schema/artifact.json', './build/schema.json');
+  } catch (e) {
+    console.error('Error writing artifact to file:', e);
+  }
```
josephgregoryii, line 109: Would we want to error out if we c ant write to artifact file?  I'd imagine we'd want some indication of that so if the artifact build fails we can have some kind of notification

d3caf: Oh yeah, I suppose we could `exit(1)` here. The actual workflow will fail if there are no files to release so we're protected in that regard.

.github/workflows/release-artifact.yml
```diff
@@ -0,0 +1,51 @@
+#  This workflow runs when a tag is created with the `v` prefix or a push
+#  to main occurs. It runs our validation script, outputs the artifact,
+#  creates a new tag, Github release, and uploads/associates the artifact with the release.
+
+name: Release artifact
+on:
+  workflow_call:
```
d3caf, line 7: DELETE THIS

.github/actions/build-validate-artifact/action.yml
```diff
@@ -1,12 +1,18 @@
 name: Build and Validate Artifact
 description: Reusable action to build and validate the component artifact
+inputs:
+  command-args:
```
d3caf, line 4: Let's switch this to be an `output-artifact` input instead of taking any args

.github/actions/build-validate-artifact/action.yml
```diff
@@ -1,12 +1,18 @@
...
   using: composite
   steps:
-    - name: Setup workspace
-      uses: "./.github/actions/bootstrap"
     - name: Build & Validate Artifact
       shell: bash
       id: build-validate-artifact
       run: |
-        cd utils && yarn build-validate-quickstart-artifact
+        cd utils && yarn build-validate-quickstart-artifact ${{ inputs.output-artifact && '--output-artifact' || '' }}
```
josephgregoryii, line 14: silly question, does this actually supply `--output-artifact`?  if `inputs.output-artifact` is truthy?

in other words, would want a ternary here?

josephgregoryii (approved):
Reviewed in a campfire, everything looked good!

We noticed in the campfire that `dataSourceIds` need to be a part of the schema and not the artifact. Mentioning this here for historical context.

.github/actions/build-validate-artifact/action.yml
d3caf: Fun fact! You can't ternary in gh actions syntax :)

PR #2550, opened 2024-09-11
feat: NR-305852 Check for missing quickstart IDs

mickeyryan42:
https://new-relic.atlassian.net/browse/NR-305852

# Summary

The job runs conditionally in PRs when the base is `main` and the head ref is `release` and validates that all quickstarts have IDs.  Also removing the requirement for the `id` on quickstarts because it isn't required at PR time and they're generated later.   We cant remove the id entirely because then when we validate on the main branch it'll fail, so i've just made them optional.

### DIFF
--- a/utils/schema/artifact.json
+++ b/utils/schema/artifact.json
@@ -79,7 +79,6 @@
         }
       },
       "required": [
-        "id",
         "description",
         "summary",
         "title",


.github/workflows/pr-checks.yml
```diff
@@ -15,3 +15,16 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332
+        with:
+          ref: ${{ github.sha }}
+      - name: Setup workspace
+        uses: "./.github/actions/bootstrap"
+      - name: Validate Quickstart IDs
+        run: |
+          cd utils && yarn validate-quickstart-ids.ts
```
d3caf, line 30: I think we missed adding the script to `package.json`. We'd need to add it and then call `cd utils && yarn validate-quickstart-ids`.

mickeyryan42: oops, thought i did that yesterday.  fixed

d3caf (dismissed):
Tested it out e2e on my fork and worked as expected 👍

d3caf:
Question: is this actually redundant, though? The schema validation ensures that quickstarts have IDs already.

josephgregoryii:
> Question: is this actually redundant, though? The schema validation ensures that quickstarts have IDs already.

This made me think, what if its a new quickstart being created? We don't have an ID for that quickstart yet

mickeyryan42:
> > Question: is this actually redundant, though? The schema validation ensures that quickstarts have IDs already.
> 
> This made me think, what if its a new quickstart being created? We don't have an ID for that quickstart yet

We have a workflow that runs on merge to the release branch, creates IDs and commits them back to the branch.  So maybe we need to think about excluding that validation from this branch or something, and only validate on merge to main?

d3caf:
Update after standup: we're not going to require `id` on quickstarts since this is auto-generated by another workflow. The job being added in this PR will ensure that all quickstarts have IDs assigned when we merge from release -> main.

d3caf (approved):

josephgregoryii (approved):

PR #2555, opened 2024-09-13
Refactor and clean up actions/worfklows

d3caf:
With the changes we've made to our validation/submission workflows, we no longer need many of the previous actions. This PR removes them and tries to consolidate things a bit as well.


mickeyryan42 (approved):
So awesome how many of these we can remove.  as far as i can tell the add-commit-status action isnt used now, is that right?  I dont see a problem leaving it even if so, might be useful sometime later.

.github/workflows/validate_packs.yml
```diff
@@ -1,281 +0,0 @@
...
-  workflow_run:
-    workflows: ["Validation Gate"]
-    types:
-      - completed
-
-env:
-  COMMIT_SHA: ${{ github.event.workflow_run.head_commit.id }}
-
-jobs:
-  ensure-quickstart-ids-are-unique:
```
josephgregoryii, line 13: I would argue that we would still want this job here. We don't currently have an in-house way to verify this.

.github/workflows/validate_packs.yml
```diff
@@ -1,281 +0,0 @@
...
-        run: cd utils && yarn check-quickstart-uniqueness
-
-      - name: Add commit status
-        if: always()
-        uses: "./.github/actions/add-commit-status"
-        with:
-          statusContext: "Validation / Quickstart id are unique"
-          state: ${{ job.status }}
-
-  ensure-images-are-valid:
```
josephgregoryii, line 48: Same here

.github/workflows/validate_packs.yml
```diff
@@ -1,281 +0,0 @@
...
-        run: cd utils && yarn validate-images
-
-      - name: Add commit status
-        if: always()
-        uses: "./.github/actions/add-commit-status"
-        with:
-          statusContext: "Validation / Image count and extension compliance"
-          state: ${{ job.status }}
-
-  ensure-icons-exist:
```
josephgregoryii, line 83: and here

.github/workflows/validate_packs.yml
```diff
@@ -1,281 +0,0 @@
...
-          cd utils && yarn validate-quickstart-data-sources "$URL"
-
-      - name: Add commit status
-        if: always()
-        uses: "./.github/actions/add-commit-status"
-        with:
-          statusContext: "Validation / Data source ids exist"
-          state: ${{ job.status }}
-
-  ensure-quickstart-dashboard-names-are-unique:
```
josephgregoryii, line 208: here as well

.github/workflows/validate_packs.yml
```diff
@@ -1,281 +0,0 @@
...
-        run: cd utils && yarn check-dashboard-name-uniqueness
-
-      - name: Add commit status
-        if: always()
-        uses: "./.github/actions/add-commit-status"
-        with:
-          statusContext: "Validation / Quickstart dashboard name is unique"
-          state: ${{ job.status }}
-
-  dashboard-helper:
```
josephgregoryii, line 243: This helps sanitize dashboard JSONs for the catalog service. I think we might still want this as well

d3caf:
@josephgregoryii raining on the deletion parade 😞

mickeyryan42 (approved):
Thought I approved this yesterday!

PR #2565, opened 2024-09-19
fix: Restore datasource validation and creation workflow

d3caf:
Accidentally removed `validate_data_sources` which also included the part of the workflow which publishes data sources. Restoring that here.


d3caf:
Closing -- this actually wasn't needed.

PR #2580, opened 2024-10-03
feat: Shorten all data source descriptions to <240 char

sarahkitten:
https://new-relic.atlassian.net/browse/NR-321974

Shortens the descriptions of the listed data sources.  Many/most of these seem to be duplicates of the quickstart descriptions, so I'm not sure if I should update those to match.

### DIFF
--- a/data-sources/datazoom/config.yml
+++ b/data-sources/datazoom/config.yml
@@ -1,7 +1,7 @@
 id: datazoom
 displayName: Datazoom
 description: |
-  The Datazoom quickstart provides a fast and easy launching point into video metrics built on top of New Relic One dashboards, allowing you to gain insights about your viewers, content, ads and your platform's performance. Customize the dashboard to use only the metrics that fit your goals and easily edit metric calculations and tailor them to your needs.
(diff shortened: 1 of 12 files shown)


data-sources/datazoom/config.yml
```diff
@@ -1,7 +1,7 @@
 id: datazoom
 displayName: Datazoom
 description: |
-  The Datazoom quickstart provides a fast and easy launching point into video metrics built on top of New Relic One dashboards, allowing you to gain insights about your viewers, content, ads and your platform's performance. Customize the dashboard to use only the metrics that fit your goals and easily edit metric calculations and tailor them to your needs.
+  The Datazoom quickstart provides a fast and easy launching point into video metrics built on top of New Relic One dashboards, allowing you to gain insights about your viewers, content, ads and your platform's performance.
```
josephgregoryii, line 4: this might be a producty question, so unsure if it applies but would we want to remove/replace references to "quickstart" since these are data sources?

josephgregoryii (approved):

d3caf (approved):
thank you!

data-sources/datazoom/config.yml
sarahkitten: yeah I figure I'll leave them for now, I imagine there's a lot of that :')

PR #2583, opened 2024-10-04
Updates to artifact to conform to service structs

josephgregoryii:
## Summary

While working on the server side of the quickstart ecosystem, we noticed that there are a lot of connected structs between data sources, quickstarts, alerts and dashboards. On top of that, we actually _lose_ a lot of functionality by not going through a mutation pipeline that helps "transform" the quickstarts / components into proper structs for the database

This PR is to re-configure the artifact and schema. The schema takes shape closely to what the service and data base expect.


utils/lib/Quickstart.ts
```diff
@@ -198,6 +198,20 @@ class Quickstart {
     };
   }
 
+  public transformForArtifact(): QuickstartConfig | { authors: { name: string }[] } {
```
josephgregoryii, line 201: TODO:
- `quickstart_uuid` over `id`
- `icon_url` over `icon`
- `authors` is a list of names

utils/schema/artifact.json
```diff
@@ -32,178 +45,576 @@
...
+            },
+            "required": ["url"]
+          }
+        }
       },
-      "required": ["name", "pages"],
+      "required": ["displayName", "rawConfiguration"],
       "additionalProperties": false
+    },
+    "dataSourceIds": {
```
d3caf, line 280: Should these be committed? I feel like these get generated...

d3caf: We think this should be fine since these get overwritten on PR builds anyway.

d3caf (approved):

mickeyryan42 (approved):

PR #2648, opened 2025-05-14
Updated documentation link for postgres

sjyothi54:
Updated the documentation link of postgres as this https://docs.newrelic.com/install/postgresql/ instead of https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/postgresql/postgresql-integration/


RamanaReddy8801 (approved):
LGTM

PR #2649, opened 2025-05-14
Set term duration to a multiple (600) of sliding window (30)

pschrimpf:
# Summary

Importing the Browser Alert Conditions results in a Validation Error:
image

This is due to the following error:
image

The PR fixes that by setting the Term duration to 600 seconds (10 minutes), keeping the sliding window duration at 30 seconds.


RamanaReddy8801 (approved):
LGTM

PR #2691, opened 2025-08-14
Revert "Rename new-dashboard.json to tidbcloud.json"

praveenkatha-nr:
Reverts newrelic/newrelic-quickstarts#2690


praveenkatha-nr:
Hi @jinshayumi. 
We haven't deployed your changes yet. Could you please clarify one thing for us?

Do you intend for both dashboard templates, TiDB Cloud Dynamic Tracker and TiDB Cloud Monitoring, to be available in the dashboards marketplace? If so, each template should be placed in a separate folder. A single folder can't contain both dashboard template JSON files. Furthermore, it's ideal to include screenshots for each template within their respective folders.

